PR #326, opened 2014-12-30
only have one main property

s-block:


cla_public/apps/checker/fields.py
```diff
@@ -248,6 +248,21 @@ def remove(self, index):
...
+        for property in self.entries:
+            if property.is_main_home.data == YES:
+                if has_main_property:
+                    property.is_main_home.errors.append(
+                        self.gettext('You can only have 1 main Property'))
+                    self.errors.append(property.is_main_home.errors)
+                has_main_property = True
+
+        return len(self.errors) == 0
+
```
andyhd: Only thing about this is that it only sets errors on the second and subsequent main property fields. I think there should be an error on all main property fields if there are more than one. Eg:
```python
main_properties = filter(is_main_property, self.entries)
if len(main_properties) > 1:
    map(set_error, main_properties)
```

s-block: I did that on purpose, but agree it would probably be cleaner to add the error to all fields

andyhd:
:+1:

PR #337, opened 2015-01-06
convert translation file to utf8 line by line

s-block:


andyhd:
LGTM. IO inefficiency is mitigated by the fact that this won't be run very often.

PR #345, opened 2015-01-07
force english category name for knoweledge base lookup

s-block:


cla_public/libs/utils.py
```diff
@@ -8,3 +9,17 @@ def get_locale():
...
+    current_app.babel.locale_selector_func = None
+    current_app.babel.localeselector(lambda: locale)
+    refresh()
+
+
+def restore_locale():
+    if hasattr(current_app.babel, 'original_locale_selector_func'):
+        current_app.babel.locale_selector_func = None
+        current_app.babel.localeselector(current_app.babel.original_locale_selector_func)
+        refresh()
```
andyhd: I would use a contextmanager here:
```python
@contextlib.contextmanager
def override_locale(locale):

    def set_locale_selector_func(fn):
        current_app.babel.locale_selector_func = None
        current_app.babel.localeselector(fn)
        refresh()

    original = current_app.babel.locale_selector_func
    set_locale_selector_func(lambda: locale)
    yield
    set_locale_selector_func(original)
```
and then use it like
```python
with override_locale('en'):
    # do stuff
```

andyhd:
:+1:

PR #347, opened 2015-01-07
remove time-based threshold scenarios

axemonkey:
* remove non-monthly income scenarios


s-block:
Were all of these failing? Could just change the weekly income to monthly - is that what the spreadsheet does?

axemonkey:
I think then you'd just end up duplicating the single gross income tests from higher up the spreadsheet

PR #350, opened 2015-01-08
send null initial values depending on AboutForm values

s-block:


cla_public/apps/checker/fields.py
```diff
@@ -150,7 +154,9 @@ def process_data(self, value):
 class MoneyIntervalForm(NoCsrfForm):
     """Money amount and interval subform"""
     per_interval_value = MoneyField(validators=[Optional()])
-    interval_period = SelectField('', choices=MONEY_INTERVALS)
+    interval_period = SelectField('',
+                                  choices=MONEY_INTERVALS,
+                                  coerce=coerce_unicode_if_value)
```
andyhd: I know this is PEP8 compliant, but I don't like indentation to an arbitrary depth like this. It's much better IMO like this:
```python
interval_period = SelectField(
    '',
    choices=MONEY_INTERVALS,
    coerce=coerce_unicode_if_value)
```

cla_public/apps/checker/forms.py
```diff
@@ -28,11 +29,57 @@
...
+            field.data = 0
+        elif isinstance(field, MoneyIntervalField):
+            field.form.per_interval_value.data = 0
+            field.form.interval_period.data = 'per_month'
+        elif isinstance(field, FormField):
+            field.form = set_zero_values(field.form)
+        elif isinstance(field, FieldList):
+            field.entries = []
+    map(set_zero, form._fields.itervalues())
+    return form
```
andyhd: Hm, lots of `isinstance`, which is a bad smell. But I can only think of one other way to do this (adding a `set_zero` method to each Field class), which would be more verbose.

cla_public/apps/checker/forms.py
```diff
@@ -145,8 +192,32 @@ def value_or_zero(field, dependant_field):
...
+        recursive_dict_update(payload, properties_data)
+
+        # set savings to zero/null depending on has_savings/has_valuables
+        savings_data = SavingsForm.get_session_as_api_payload() if \
+            self.require_savings else SavingsForm.get_zero_api_payload()
+        recursive_dict_update(payload, savings_data)
+
+        # add null/session income and set to 0 if on passported benefits later
+        income = IncomeForm().get_session_as_api_payload()
+        recursive_dict_update(payload, income)
```
andyhd: Comments mean the code is not clear enough and there's repetition of `Form.get_session_as_api_payload` and `recursive_dict_update`. I think it would be clearer abstracted into a function, eg:
```python
def update_payload(form_class, cond=True):
    if cond:
        form_data = form_class.get_session_as_api_payload()
    else:
        form_data = form_class.get_zero_api_payload()
    recursive_dict_update(payload, form_data)

update_payload(PropertiesForm, cond=self.require_properties)
update_payload(SavingsForm, cond=self.require_savings)
update_payload(IncomeForm)
```

cla_public/apps/checker/forms.py
```diff
@@ -417,8 +494,17 @@ def api_payload(self):
         }
 
 
-class IncomeAndTaxForm(ConfigFormMixin, Honeypot, Form):
+class IncomeForm(ConfigFormMixin, Honeypot, Form, FormSessionDataMixin):
     your_income = FormField(IncomeFieldForm, label=_(u'Your personal income'))
+    partner_income = PassKwargsToFormField(IncomeFieldForm,
+                                           form_kwargs={'is_partner': True},
+                                           label=_(u'Your partner‘s income'))
```
andyhd: Is that the apostrophe facing the right way?

cla_public/apps/checker/forms.py
```diff
@@ -417,8 +494,17 @@ def api_payload(self):
...
+
+    def __init__(self, *args, **kwargs):
+        """Dynamically remove partner subform if user has no partner"""
+        super(IncomeForm, self).__init__(*args, **kwargs)
+        if not session.has_partner:
+            del self.partner_income
```
andyhd: This does make more sense :+1:

andyhd:
LGTM

PR #358, opened 2015-01-12
Can disable checker and only show call-me-back form by setting an env var

andyhd:


tyom:
Should we add `CALLMEBACK_ONLY` to example config file?

tyom:
Looks good to me.

andyhd:
`CALLMEBACK_ONLY` is not really meant to be changed except through the salt config for a deployment. I don't think we need to put it in the example config.

PR #362, opened 2015-01-13
add final translations

s-block:


cla_public/apps/checker/forms.py
```diff
@@ -245,8 +246,9 @@ def api_payload(self):
 
         return payload
 
+from speaklater import is_lazy_string
```
andyhd: why is this all the way down here?

s-block: not actually needed

andyhd:
LGTM

PR #376, opened 2015-01-19
Debugger: Add missing fields and option to set all fields to Yes or No

tyom:


axemonkey:
Something a little funky going on with the 'No' button.
screen shot 2015-01-20 at 10 09 08
and
screen shot 2015-01-20 at 10 09 17

tyom:
> Something a little funky going on with the 'No' button.

Ah, good catch. It was breaking when no category was selected (the Problem page). Fixed now.

axemonkey:
Superb!

PR #387, opened 2015-01-23
Distinguish between Zendesk feedback on different environments

tyom:


cla_public/apps/base/forms.py
```diff
@@ -49,15 +51,21 @@ class FeedbackForm(Honeypot, BabelTranslationsFormMixin, Form):
         validators=[InputRequired()])
 
     def api_payload(self):
+        environment = os.environ.get('CLA_ENV')
```
s-block: could add this to the main settings file

CLA_ENV = os.environ.get('CLA_ENV')

then

environment = current_app.config['CLA_ENV']

s-block:
Looks good

tyom:
Updated.

s-block:
Looks good

PR #391, opened 2015-01-27
IE currency bug fix + CSS fix

axemonkey:
* fix currency formatting bug in all IE versions
* fix blue legends in IE7


tyom:
LGTM

tyom:
Looks good.

PR #394, opened 2015-01-27
Send to GA eligibility status from confirmation page

tyom:


cla_public/apps/callmeback/views.py
```diff
@@ -46,6 +46,7 @@ def request_callback():
 @callmeback.route('/result/confirmation')
 @redirect_if_no_session()
 def confirmation():
-    response = render_template('result/confirmation.html')
+    response = render_template('result/confirmation.html',
+        eligibility_status=session.eligibility_status)
```
andyhd: This value is already available in the template via `{{ session.eligibility_status }}`, there's no need to pass it explicitly.

cla_public/apps/checker/session.py
```diff
@@ -145,6 +145,20 @@ def aged_60_or_over(self):
...
+    @property
+    def eligibility_status(self):
+        status = 'unknown'
+        is_eligible = self.get('is_eligible', ELIGIBILITY_STATES.UNKNOWN)
+
+        if is_eligible == ELIGIBILITY_STATES.YES:
+            status = 'eligible'
+        elif is_eligible == ELIGIBILITY_STATES.NO:
+            status = 'ineligible'
+        elif is_eligible == ELIGIBILITY_STATES.UNKNOWN:
```
andyhd: This branch is redundant

cla_public/apps/checker/session.py
```diff
@@ -145,6 +145,20 @@ def aged_60_or_over(self):
     def callback_time(self):
         return self.get('CallMeBackForm_time')
 
+    @property
+    def eligibility_status(self):
+        status = 'unknown'
+        is_eligible = self.get('is_eligible', ELIGIBILITY_STATES.UNKNOWN)
```
andyhd: The default value is not really needed here either.

cla_public/apps/callmeback/views.py
tyom: The reason it's passed explicitly is because the session is cleared on the following line.

andyhd: The template is rendered before the session is cleared. The session var is available in the template.

cla_public/apps/checker/session.py
tyom: Ah, the initial `status` should be empty string. The idea is that the status only becomes unknown when the at least one form is submitted. Unfortunately this includes the details form, which I'm hoping can be fixed separately in near future.

cla_public/apps/checker/session.py
tyom: Currently `is_eligible` is set only after completing “About you” form. Therefore `eligibility_status` will throw when it's accessed from pages before to completing the “About you” form (i.e. going straight to “Call me back” page).

cla_public/apps/callmeback/views.py
tyom: You're right. Updating.

andyhd:
:+1:

PR #401, opened 2015-02-03
don't run threshold tests locally any more

axemonkey:


axemonkey:
retest this please

andyhd:
:+1:

tyom:
Should we remove `threshold.js` if we're not using it any more?

axemonkey:
Josh seemed keen on keeping it in case we wanted it again in the future. I'm fine with losing it as well as all the threshold json files, as they'll be in the github history.

tyom:
> as they'll be in the github history.

That's what I thought.

s-block:
I have still been running them as the unit tests have changed a lot

axemonkey:
Sooo...?

tyom:
@s-block Do you still use them?

tyom:
Discussed off-line. The consensus seems to be to move Threshold test to a different directory ([group](http://nightwatchjs.org/guide#runner-options)) and exclude by default. Perhaps adding a command to run all tests.

PR #418, opened 2015-02-13
temporary JS to route to scope checker prototype for user testing

axemonkey:
* Double-click BETA tag to enable testing mode on problem page
* Select one of the categories in the scope of the scope checker (debt, discrimination, education, family, housing, violence, immigration)
* The prototype scope checker (on AWS S3) will be opened in front of the application for testing
* The application will continue to the means test in the original tab
* When the scope checker ends at a positive result, the user clicks 'Continue' in the prototype and that window closes, revealing the means test
* When the scope checker ends at a negative result, that is displayed in the prototype. The tester can then close that window and return the original application to the start manually

Will only apply on localhost or on demo environment.

### DIFF
--- a/cla_public/static-src/javascripts/modules/user-testing.js
+++ b/cla_public/static-src/javascripts/modules/user-testing.js
@@ -1,0 +1,131 @@
+(function () {
+  'use strict';
+
+  moj.Modules.UserTesting = {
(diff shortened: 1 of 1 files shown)


tyom:
Cool.

PR #419, opened 2015-02-19
Call me back split

tyom:
[PR](https://www.pivotaltracker.com/story/show/87800282)

### DIFF
--- a/cla_public/apps/callmeback/views.py
+++ b/cla_public/apps/callmeback/views.py
@@ -3,7 +3,7 @@
 
 import logging
 
-from flask import redirect, render_template, session, url_for, views
+from flask import abort, redirect, render_template, session, url_for, views
 from flask.ext.babel import lazy_gettext as _
 from cla_public.libs.utils import log_to_sentry
 from slumber.exceptions import SlumberBaseException
@@ -61,13 +61,13 @@ callmeback.add_url_rule(
     view_func=CallMeBack.as_view('request_callback'))
 
 
-class CallMeBackConfirmation(views.MethodView, object):
+class CallMeBackConfirmation(views.MethodView):
 
     def get(self):
-        # render template before clearing session, as we use it in the template
-        response = render_template('result/confirmation.html')
-        session.clear()
-        return response
+        session.clear_and_store_ref()
+        if not session.get('stored_case_ref'):
+            abort(404)
+        return render_template('result/confirmation.html')
 
 
 callmeback.add_url_rule(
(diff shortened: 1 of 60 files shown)


s-block:
retest this please

s-block:
retest this please

s-block:
retest this please

cla_public/apps/checker/session.py
```diff
@@ -182,6 +182,22 @@ def api_payload(self):
...
+        if not self.get('is_expired', False):
+            store = {
+                'is_expired': True,
+                'stored_case_ref': self.get('case_ref'),
+                'stored_callback_time': self.callback_time,
+                'stored_callback_requested': self.callback_requested,
+                'stored_category': self.category,
+            }
+            self.clear()
+            self.update(store)
```
andyhd: Why do this instead of rendering the template and then clearing the session? This is more complicated and now there is data in the session after the user has completed the journey.
Also, why use 'stored_' attributes when the originals have been removed? They could just be re-added.

cla_public/apps/callmeback/views.py
```diff
@@ -57,17 +57,17 @@ def dispatch_request(self, *args, **kwargs):
...
 
     def get(self):
-        # render template before clearing session, as we use it in the template
-        response = render_template('result/confirmation.html')
-        session.clear()
-        return response
+        session.clear_and_store_ref()
+        if not session.get('stored_case_ref'):
+            abort(404)
+        return render_template('result/confirmation.html')
```
andyhd: There's no need for `session.clear_and_store_ref`:
```python
if not session.get('case_ref'):
    abort(404)
response = render_template('result/confirmation.html')
session.clear()
return response
```

cla_public/libs/views.py
```diff
@@ -19,7 +19,7 @@ class RequiresSession(object):
     session_expired_url = '/session-expired'
 
     def dispatch_request(self, *args, **kwargs):
-        if not session:
+        if not session or session.get('is_expired', False):
```
andyhd: Not necessary if session is cleared

cla_public/apps/callmeback/views.py
tyom: Discussed offline and explained the need for that method.

axemonkey:
LGTM from a FE perspective

s-block:
Looks good

andyhd:
:+1:

tyom:
retest this please

tyom:
retest this please

PR #423, opened 2015-02-24
Add spec paths to Makefile

tyom:
For individual tests use:

`make test spec=spec-file-name` (without extension)


axemonkey:
Sure

PR #435, opened 2015-03-17
run tests on browserstack

axemonkey:
* run tests on browserstack
* fix a few tests to work in IE8 (no :last-child)
* increase time allowed when waiting for elements
* tweak validation testing to not resubmit form to check for every single error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,19 @@ ifdef spec
 	specific_test = -t ${SPECS_PATH}/${spec}.js
 endif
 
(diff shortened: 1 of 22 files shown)


andyhd:
LGTM - the only thing I question is why we have a Makefile in addition to a Procfile.dev, gulpfile.js, manage.py, scripts/, tasks/ and tools/ :smile:

axemonkey:
That would be a Tyom question - he created the Makefile I believe. :grinning:

PR #436, opened 2015-03-17
Improvements for conditional fields

tyom:
Add wildcard support to trigger element visibility through `data-controlled-by` when any value is `data-control-value` is selected.


axemonkey:
LGTM

PR #442, opened 2015-04-01
Fix callback times in BST

andyhd:


cla_public/apps/contact/fields.py
```diff
@@ -194,7 +194,7 @@ def data(self):
         return self.scheduled_time()
 
     def scheduled_time(self):
-        return self.form.scheduled_time().replace(tzinfo=pytz.utc)
+        return self.form.scheduled_time()
```
tyom: Should you remove `import pytz` above now that it's not used?

tyom:
LGTM

PR #444, opened 2015-04-08
add new disregarded benefit to constants

axemonkey:
A new disregarded benefit should be added to the site from April 13th.

[Pivotal story](https://www.pivotaltracker.com/story/show/91280458)


tyom:
Sure.

PR #449, opened 2015-04-10
Make the MoneyIntervalRequiredValidator stop further validation if there is no

r4vi:
value there and the field is required

this is because a user putting in a blank value but choosing a interval period
is currently getting two errors


andyhd:
:metal:

r4vi:
retest this please

r4vi:
:facepunch: :family: :+1:

PR #452, opened 2015-04-13
keep the selected time/day set by the client if they get a validation er...

r4vi:
...ror

this was a bug caused by the way SelectField in wtf works, it expects the format of self.data to match self.choices.

### DIFF
--- a/cla_public/apps/contact/fields.py
+++ b/cla_public/apps/contact/fields.py
@@ -36,6 +36,10 @@ class FormattedChoiceField(object):
         except ValueError:
             self.data = value
 
+    def iter_choices(self):
+        for value, label in self.choices:
+            yield (value, label, self.coerce(value) == self._format(self.data))
+
     def pre_validate(self, form):
         choice_values = (v for v, _ in self.choices)
         if self._format(self.data) not in choice_values:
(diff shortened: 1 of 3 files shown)


axemonkey:
FE LGTM. :+1:

s-block:
LGTM

r4vi:
:rocket:

cla_public/apps/contact/fields.py
```diff
@@ -76,7 +80,9 @@ def process_formdata(self, valuelist):
 
     @classmethod
     def _format(cls, value):
-        return '{:%Y%m%d}'.format(value)
+        if value and not isinstance(value, basestring):
```
andyhd: if you're going to do `isinstance` here, why not make it `isinstance(value, datetime.datetime)` (or `datetime.date`)

r4vi: wanted to keep FormattedChoiceField generic enough so that if someone wanted to use it for something else that needs formatting they can just implement a get_format_string method. (not assumed to be a format string for a datetime) and if it's already a string then don't do anything.

andyhd: I didn't see which class this method was in. But it's already assuming the value is a date with the hardcoded format string.

andyhd: No wait, bloody folds. It's in `DayChoiceField`, so it's definitely a datetime or date

r4vi: I know but a good refactor could be to move ._format into FormattedChoiceField. then add FormattedChoiceField.get_format_string on which looks at self.FORMAT_STRING then nothing is assumed by FormattedChoiceField about types and we can move this very similar looking fn out of DayChoiceField and TimeChoiceField

andyhd: Yep, agreed

PR #461, opened 2015-04-23
Send case ref by email

tyom:
- [x] Add Sendgrid ENV variables to all environments


tyom:
test this please

cla_public/apps/contact/views.py
```diff
@@ -37,6 +66,9 @@ def on_valid_submit(self):
                 u'Please check and try again.')
             return self.get()
         else:
+            if self.form['email'].data:
```
andyhd: Shouldn't this be `self.form.email.data`?

cla_public/libs/mailgun.py
```diff
@@ -0,0 +1,22 @@
+import requests
+from flask import current_app
+
+
+def send_ref_by_email(to, subject, email_body):
```
andyhd: I think you should rename this function - there's no mention of refs in the function.
Something like `send_email` would be better.

cla_public/libs/mailgun.py
```diff
@@ -0,0 +1,22 @@
...
+    return requests.post(
+        'https://api.mailgun.net/v3/{domain}/messages'.format(
```
andyhd: This URL should be in config

cla_public/apps/contact/views.py
```diff
@@ -20,6 +22,33 @@ def add_no_cache_headers(response):
     return response
 
 
+def handle_confirmation_email(form):
+    data = {
+        'case_ref': session['case_ref'],
+        'full_name': form['full_name'].data,
```
andyhd: Shouldn't all of these `form['something']`s be `form.something`?

cla_public/apps/contact/forms.py
```diff
@@ -136,6 +136,15 @@ class ContactForm(Honeypot, BabelTranslationsFormMixin, Form):
         validators=[
             IgnoreIf('callback_requested', FieldValue(NO))
         ])
+    email = StringField(
+        _(u'Email address'),
+        description=_(u'If you add your email we will send you the '
+                      u'reference number when you submit your details'),
```
andyhd: This is better formatted like this:
```python
description=_(
    u'If you add ...'
    u'reference ...'),
```
because the indentation level would be less arbitrary.

cla_public/libs/mailgun.py
tyom: Config or local constant like [in Zendesk](https://github.com/ministryofjustice/cla_public/blob/develop/cla_public/libs/zendesk.py#L9)?

andyhd: Config - I think the Zendesk URL should be in config too.

tyom: Isn't it specific to this implementation though?

andyhd: I don't imagine the API will change much, but the version number in the URL might. I guess it doesn't matter that much. But most of our API URLs are in config - it seems better to keep them together if you need to find them.

tyom: Resolved off-line. Leaving as-is due to added complexity moving them to constants and referring other constants inside `format`.

cla_public/apps/contact/forms.py
```diff
@@ -136,6 +136,17 @@ class ContactForm(Honeypot, BabelTranslationsFormMixin, Form):
...
         ])
+    email = StringField(
+        _(u'Email address'),
+        description=_(
+            u'If you add your email we will send you the '
+            u'reference number when you submit your details'
+        ),
+        validators=[
+            Length(max=255, message=_(u'Your address must be 255 characters '
+                                      u'or less')),
```
andyhd: Indentation here should be as on line 142

andyhd:
:+1:

andyhd:
:+1:

andyhd:
retest this please

PR #466, opened 2015-04-24
Feature/ineligible reason

s-block:


cla_public/templates/help-organisations.html
```diff
@@ -14,6 +14,18 @@
     {% if category == 'violence' %} {{ _('for the advice they can provide') }}{% endif %}.
   </p>
 
+  {% if ELIGIBILITY_REASONS.DISPOSABLE_CAPITAL in ineligible_reasons %}
+    <h3>DISPOSABLE_CAPITAL</h3>
```
andyhd: This doesn't look complete

cla_public/apps/checker/constants.py
```diff
@@ -174,3 +176,18 @@
     'publiclaw': 'pub',
     'benefits': 'wb'
 }
+
+ELIGIBILITY_REASONS_TEXT = {
```
andyhd: I think copy should be in the template, not in the Python code

andyhd:
:+1:

PR #473, opened 2015-05-01
Implement textarea character counter

tyom:
Usage:
Add `data-character-count='4000'` as attribute on textarea.

The value has to be a number. A small template for counter is also required (see contact.html)

[#92355174]


axemonkey:
Looks good.

PR #483, opened 2015-05-15
update copy for confirmation page in DV cases

axemonkey:
[Pivotal](https://www.pivotaltracker.com/story/show/94508636)


tyom:
:+1:

PR #491, opened 2015-05-26
why are users dropping out / reasons for contacting

ushkarev:
Interstitial form to determine why users are dropping out of the process submitting to Zendesk

### DIFF
--- a/tests/nightwatch/specs/current/contact-page.js
+++ b/tests/nightwatch/specs/current/contact-page.js
@@ -7,8 +7,14 @@ function text(n) {
 function contactPage(client) {
   client
     .startService()
-    .click('#callback-link')
-  ;
+    .click('#callback-link', function() {
+      console.log('     ⟡ "Get in touch" link clicked');
+    })
+    .waitForElementVisible('.reasons-for-contacting-form', 3000,
+      '  - "Reasons for contacting" form exists')
+    .submitForm('.reasons-for-contacting-form', function() {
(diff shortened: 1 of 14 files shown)


tyom:
I think 'Other reasons' should be a subfield under 'Another reason' button, particularly that it expands it. Having it outside of bordered fieldset seems odd.

screen shot 2015-05-29 at 15 47 25

vs 

screen shot 2015-05-29 at 15 48 32

ushkarev:
yeah i thought of that too, but then i can't reuse the macros used everywhere else

tests/nightwatch/specs/current/contact-page.js
```diff
@@ -7,8 +7,14 @@ function text(n) {
...
     .startService()
-    .click('#callback-link')
-  ;
+    .click('#callback-link', function() {
+      console.log('     ◇ "Get in touch" link clicked');
+    })
+    .waitForElementVisible('.reasons-for-contacting-form', 3000,
+      '  - "Reasons for contacting" form exists')
+    .submitForm('.reasons-for-contacting-form', function() {
+      console.log('     ◇ "Continue to contact CLA" button clicked');
```
tyom: Can I suggest using `⟡` as I updated that character in [another PR](https://github.com/ministryofjustice/cla_public/pull/496/files#diff-7ce2dbe1781eb931b65932e83ea02e0eR22). I think it works better with other terminal fonts.

tyom:
LGTMO

ushkarev:
retest this please

PR #492, opened 2015-05-26
support for sending current locale to backend API

ushkarev:
for getting Welsh text choices from the GraphML


cla_public/apps/contact/forms.py
```diff
@@ -45,6 +46,13 @@ class AdaptationsForm(BabelTranslationsFormMixin, NoCsrfForm):
         _(u'Other communication needs'),
         description=_(u'Please tell us what you need in the box below'))
 
+    def __init__(self, formdata=None, obj=None, prefix='', data=None, meta=None, **kwargs):
+        if data is None:
```
s-block: maybe should check to see if formdata is None as well, also maybe ojb

s-block:
:+1:

PR #493, opened 2015-05-26
contact page was not reachable without session

ushkarev:
but it's linked from the home page where no session existed


tyom:
I wonder if contact page should not require session at all, so it can be linked to directly from anywhere?

ushkarev:
i agree that'd be better, but this is what josh suggested. there's something the contact form checks for in the session

s-block:
:+1:

PR #494, opened 2015-05-26
Feature/form errors

s-block:


cla_public/templates/base.html
```diff
@@ -68,14 +68,35 @@
...
       {{ _('Close') }}
     </button>
   </script>
+
+  <script type="text/html" id="mainFormError">
+    <div class="alert alert-error" role="alert">
+      <span class="icon icon-cross" aria-hidden="true"></span>
+      <div class="alert-message">
+        <p>This form has errors.</p>
+        <p>Please see below for the errors you need to correct.</p>
```
ushkarev: shouldn't this bit use translation tags?

s-block: correct. will do.

tyom:
Excellent work! :+1:

tyom:
retest this please

PR #497, opened 2015-05-28
show postcode message as help, not error

axemonkey:
[Pivotal](https://www.pivotaltracker.com/story/show/93978358)

### DIFF
--- a/cla_public/static-src/javascripts/modules/address-finder.js
+++ b/cla_public/static-src/javascripts/modules/address-finder.js
@@ -73,7 +73,7 @@
       this.addresses = data;
 
       if (data.length < 1) {
-        this.showError('No addresses were found with that postcode');
+        this.showHelp('No addresses were found with that postcode, but you can still enter your address manually');
       } else if (data.length === 1) {
         this.updatePostcode(data);
         this.updateAddress(0);
@@ -132,6 +132,11 @@
         .addClass('m-error');
       this.$formGroup.find('.form-row')
         .before($('<div class="form-row field-error"><p>' + msg + '</p></div>'));
+    },
+
+    showHelp: function (msg) {
+      this.$formGroup.find('.form-row')
+        .before($('<div class="form-row field-help"><p>' + msg + '</p></div>'));
     },
 
     reset: function () {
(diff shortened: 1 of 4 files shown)


cla_public/static-src/javascripts/modules/address-finder.js
```diff
@@ -73,7 +73,7 @@
       this.addresses = data;
 
       if (data.length < 1) {
-        this.showError('No addresses were found with that postcode');
+        this.showHelp('No addresses were found with that postcode, but you can still enter your address manually');
```
ushkarev: shouldn't this be welsh-ified somehow? do we have a mechanism to get welsh into js?

axemonkey: There are a number of these in JS, and it doesn't look like they are being translated at the moment.

But yes, they probably should. :-)

axemonkey: Yes probably. :-)

axemonkey: OK there are three lines of content in that JS file. We should probably use underscore templates, as with other translated elements on the same page. I'll do that tomorrow.

tyom: I'd suggest using a different style to help text. A warning style (yellow maybe)?

tyom:
The same should be for 'must contain a valid postcode' "error".

cla_public/static-src/javascripts/modules/address-finder.js
```diff
@@ -134,6 +134,11 @@
         .before($('<div class="form-row field-error"><p>' + msg + '</p></div>'));
     },
 
+    showHelp: function (msg) {
+      this.$formGroup.find('.form-row')
```
tyom: You need to clear it before adding. Otherwise:

screen shot 2015-05-28 at 17 29 34

tyom:
I wonder if we should just leave the error message styled as it currently is (red) and just remove the `.m-error` from `.fieldset-label`. Just getting rid of that error cross would suffice, I think.

screen shot 2015-05-28 at 18 21 25

tyom:
:+1: :shipit:

PR #501, opened 2015-06-02
Fix Scope diagnosis smoke test

tyom:


ushkarev:
looks good. the only thing i'd say is it'd be nice to put in comments to describe the steps you're following in case they change order (e.g. debt > homeless)

tyom:
Updated to extend test.

PR #504, opened 2015-06-05
email tests failed when run too late in the day

ushkarev:
because no more operator callback slots would be available


tyom:
:+1:

tyom:
test this please

PR #510, opened 2015-06-17
missing translations

ushkarev:
untranslated strings marked for translation and some sentences combined into single messages
(because sentence fragments cannot always be translated when split up)

### DIFF
--- a/cla_public/templates/contact.html
+++ b/cla_public/templates/contact.html
@@ -119,8 +119,8 @@
   <script type="text/html" id="addressFinderList">
     <div class="form-group form-subfield address-list">
       <select name="address" class="form-control">
-        <option value="" disabled><%= count %> addresses found</option>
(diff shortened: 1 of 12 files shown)


cla_public/templates/contact.html
```diff
@@ -119,8 +119,8 @@ <h2 class="subtitle">
   <script type="text/html" id="addressFinderList">
     <div class="form-group form-subfield address-list">
       <select name="address" class="form-control">
-        <option value="" disabled><%= count %> addresses found</option>
-        <option value="">-- Choose address --</option>
+        <option value="" disabled><%= count %> {{ _('addresses found') }}</option>
+        <option value="">{{ _('-- Choose address --') }}</option>
```
s-block: '--' doesn't really need to be in translation

s-block:
:+1:

PR #511, opened 2015-06-17
progress bar fix

ushkarev:
sometimes when navigating around, perhaps after session expires, users could land on a page that was unavailable in their flow. this is a quick fix to prevent the progress bar causing a 500. a full fix would redirect users to a different page (e.g. start or review)


tyom:
:+1:

PR #518, opened 2015-06-25
Feature/bank holiday fix

s-block:


ushkarev:
:+1:
tho if we get any more rules in here it feels that having an editable calendar in the DB might be easier…

PR #527, opened 2015-07-03
Categories fixes

tyom:


tyom:
test this please

cla_public/apps/checker/views.py
```diff
@@ -363,7 +363,7 @@ def get_context(self, category_name):
             name = unicode(name)
 
             category_name = ORGANISATION_CATEGORY_MAPPING.get(name, name)
-        trans_category_name = ORGANISATION_CATEGORY_MAPPING.get(name, name)
+        trans_category_name = next(c[1] for c in CATEGORIES if c[0] == category)
```
ushkarev: could `category` in this line ever be missing from `CATEGORIES`? if so, then this would throw an error

ushkarev:
:+1:

cla_public/apps/checker/views.py
tyom: Lines 358-361 should cover that case.

tyom:
test this please

PR #547, opened 2015-07-17
showing security message about keeping data safe

ushkarev:
depends on [cla_backend#378](https://github.com/ministryofjustice/cla_backend/pull/378)


tyom:
:+1:

PR #557, opened 2015-07-28
Fix bug where app is crashing when referrer is not set

tyom:


s-block:
:+1:

tyom:
test this please

PR #565, opened 2015-08-05
Improvements in accessibility

tyom:
[PT#100105536](https://www.pivotaltracker.com/story/show/100105536)


ushkarev:
i'd suggest a change to https://github.com/ministryofjustice/cla_public/commit/face5e138988ba4c085e6d76aa62d4ffdae54c20. the macro `Element.link_new_window('url', 'some text', True)` should mark up the link text for translation in line with the macro call rather than in the macro itself. otherwise the messages cannot be extracted into the .pot file for translation

ushkarev:
otherwise :+1:

tyom:
@ushkarev done.

PR #569, opened 2015-08-11
jenkins job fix: nightwatch fails to clean up these process

ushkarev:
NB: if two jobs are running on the same jenkins slave then one may break the other


s-block:
:+1:

PR #570, opened 2015-08-12
handle laalaa search errors and unavailability better

ushkarev:


cla_public/apps/checker/views.py
```diff
@@ -52,9 +52,13 @@ def handle_find_legal_adviser_form(form, args):
...
                 page = args['page']
-            data = laalaa.find(args['postcode'], category, page)
-            if 'error' in data:
-                form.postcode.errors.append(data['error'])
+            try:
+                data = laalaa.find(args['postcode'], category, page)
+                if 'error' in data:
+                    form.postcode.errors.append(data['error'])
+            except laalaa.LaaLaaError:
+                form.postcode.errors.append(_('Error looking up legal advisers.') + ' ' +
```
s-block: shouldn't concatenate lazy strings and strings. should do something like
```python
form.postcode.errors.append(u'%s %s' % (_('Error looking up legal advisers.'),
                                                                    _('Please try again later.')))
```

PR #582, opened 2015-09-28
immigration scope

ushkarev:
depends on [cla_backend#405](https://github.com/ministryofjustice/cla_backend/pull/405)


tyom:
:+1:

PR #585, opened 2015-11-27
Fix multiple properties JS bug

tyom:
When property is injected via XHR it loses form field bindings (such as conditional fields, label select etc)
Reinitialise all moj modules when the page is updated.


s-block:
:8ball:

s-block:
:+1:

PR #588, opened 2015-12-04
Fix map/results rounding issue on FALA when zooming out page

jsugarman:


s-block:
Totally professional :+1: . A+. Keep up the good work.

PR #596, opened 2016-04-08
Add ping.json

jimnarey:


s-block:

s-block:

s-block:
add to whitelist

Dockerfile
```diff
@@ -10,6 +10,7 @@ MAINTAINER Peter Idah <peter.idah@digital.justice.gov.uk>
 
 # Set correct environment variables.
 ENV HOME /root
+ENV BACKEND_BASE_URI http://ip-172-31-27-7.eu-west-1.compute.internal
```
s-block: Don't set these env vars in the docker file. It is already set from the deploy repo and should be the actual backend url not the internal address

cla_public/apps/base/views.py
```diff
@@ -18,6 +20,10 @@
 from cla_public.libs import zendesk
 from cla_public.libs.views import HasFormMixin, ValidFormOnOptions
 
+from moj_irat.views import HealthcheckView
```
s-block: this doesn't seem to be used and it is a django package

cla_public/apps/base/views.py
```diff
@@ -18,6 +20,10 @@
...
+
+from django.conf import settings
```
s-block: flask app has config in ```current_app.config```

cla_public/config/common.py
```diff
@@ -11,6 +11,14 @@
 
 CLEAR_SESSION = True
 
+HEALTHCHECKS = [
+    'moj_irat.healthchecks.database_healthcheck',
+]
+
+AUTODISCOVER_HEALTHCHECKS = True
```
s-block: these are settings for moj_irat which isn't used

cla_public/apps/base/views.py
```diff
@@ -234,3 +240,20 @@ def smoke_tests():
...
+        'build_date': os.environ.get('APP_BUILD_DATE'),
+        'commit_id': os.environ.get('APP_GIT_COMMIT'),
+        'build_tag': os.environ.get('APP_BUILD_TAG')
+    })
+
+
+@base.route('/healthcheck.json')
+def healthcheck():
+    backend_healthcheck_uri = '%s/%s' % (current_app.config['BACKEND_BASE_URI'], 'status/healthcheck.json')
+    response = requests.get(backend_healthcheck_uri)
```
s-block: should just test the response code is OK and return its own healthcheck rather than the backend one

requirements/base.txt
```diff
@@ -36,3 +36,6 @@ git+git://github.com/ministryofjustice/slumber.git@da38360d6d158ee7ed445b43228fe
 
 # Fork of WTForms with welsh translations - PR added
 git+git://github.com/s-block/wtforms.git@47f9f6d789cdb3eef3bf659e047f1bbd1f42af24
+
+#Irat healthcheck and ping package
+git+https://github.com/ministryofjustice/django-moj-irat.git
```
s-block: This isn't used and is a django package

Dockerfile
```diff
@@ -10,6 +10,7 @@ MAINTAINER Peter Idah <peter.idah@digital.justice.gov.uk>
 
 # Set correct environment variables.
 ENV HOME /root
+
```
s-block: can just revert this file. has added line in it

requirements/base.txt
```diff
@@ -36,3 +36,4 @@ git+git://github.com/ministryofjustice/slumber.git@da38360d6d158ee7ed445b43228fe
 
 # Fork of WTForms with welsh translations - PR added
 git+git://github.com/s-block/wtforms.git@47f9f6d789cdb3eef3bf659e047f1bbd1f42af24
+
```
s-block: can revert this file. has new line added

PR #597, opened 2016-04-21
Fix Sentry errors

s-block:
- if backend returns case already exists error then retrieve the case ref form the backend and redirect to confirmation without re-saving case


jimnarey:
Looks good.

PR #601, opened 2016-06-28
do not allow commas in email address

s-block:


cla_public/apps/checker/tests/test_validation.py
```diff
@@ -63,7 +63,10 @@ def test_email_validator(self):
         field.data = 'emailwithspace@example.com '
         self.assertRaises(ValidationError, validator, form, field)
 
-        field.data = 'emailwithspace@example.com'
+        field.data = 'emailwith,comma@example.com'
+        self.assertRaises(ValidationError, validator, form, field)
+
+        field.data = 'email.with-space@example.com'
```
cliffxuan: `email.with-hyphen`?

s-block: yeah a hyphen is allowed

cliffxuan: yes, the fixture should be `field.data = 'email.with-hyphen@example.com'` instead of `'email.with-space@example.com'`

PR #609, opened 2017-03-07
Fix environment check

cliffxuan:
Was always returning false and consequently incorrectly served develop branch of remote config file for prod environment


s-block:

cliffxuan:
this is a fix created by @solidgoldpig and i can verify it's needed for `master`.

PR #611, opened 2017-03-29
Check if cait_journey exists before trying to use it

solidgoldpig:


s-block:

s-block:

PR #612, opened 2017-12-04
Add app version standard env support

niallcreech:
Add missing dependencies and upgrade pyopenssl

Django version needs to be pinned < 2.0.0 to avoid a major version
upgrade. Additionally, pyopenssl needs to be upgraded to avoid an error
on no attribute 'SSL_ST_INIT'


s-block:

s-block:

PR #705, opened 2018-07-03
Move unit tests and Docker build to CircleCI

sldblog:

- Adds CircleCI job for Python unit tests.
- Adds CircleCI job for building the `cla_public` Docker image.

As a follow-up, we should remove both of these functions from our self-hosted Jenkins instance.

The Docker images are assumed to be stored in Amazon Elastic Container Registry. This might change in the future to Docker Hub or quay.io.

### Post-merge steps

- [ ] Disable previous Jenkins job that was building the Docker image
- [ ] Deployment job in Jenkins modified to use the new Docker registry

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,18 @@
 version: 2
 jobs:
+  build:
+    docker:
+      - image: docker:18-git
+    environment:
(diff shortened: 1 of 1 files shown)


sldblog:
:bowing_man:

PR #709, opened 2018-07-20
LGA-66 Add link to user survey on access to justice

rowlando:


.gitignore
```diff
@@ -59,6 +59,9 @@ venv
 *.iml
 *.iws
 
+# Visual Studio Code
+.vscode
```
sldblog: _nitpick_
I think these kinds of development environment specific ignores could go to everyone's global `.gitignore`. What do you think?

cla_public/templates/_phase-tag.html
```diff
@@ -10,10 +10,25 @@
       <a lang="cy" href="{{ url_for('base.set_locale', locale='cy_GB') }}" data-ga="event:locale/switch/welsh">Cymraeg</a>
     {% endif %}
   </div>
-
+  {#
```
sldblog: Could we remove the original block altogether? We can revert this PR to restore it if we need to.

.gitignore
rowlando: Oh yeah, forgot about. I've updated my global one.

https://github.com/rowlando/homedir/blob/master/etc/git/ignore

PR #730, opened 2018-08-06
LGA-103: Switch to previous docker registry

sldblog:

Switches the Amazon Elastic Container Registry added in #705 back to the original Docker registry.

### Why?

We need to do this as our deployment process does not currently handle other Docker registries and this is the path to least resistance on making everything build from CircleCI until we can switch the deployment infrastructure to Kubernetes.

I had to lock the remote Docker version to 17.03-ce due to the Docker registry's compatibility issues with newer versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,18 +2,19 @@ version: 2
 jobs:
   build:
     docker:
-      - image: docker:18-git
+      - image: docker:17.03-git
     environment:
       DOCKER_REGISTRY: "registry.service.dsd.io"
       DOCKER_IMAGE: "cla_public"
     steps:
       - checkout
       - setup_remote_docker:
+          version: 17.03.0-ce
           docker_layer_caching: true
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2,33 +2,35 @@ version: 2
...
-      - image: docker:18-git
+      - image: docker:17.03-git
     environment:
-      APP_NAME: "get-access/cla_public"
+      DOCKER_REGISTRY: "registry.service.dsd.io"
+      DOCKER_IMAGE: "cla_public"
     steps:
       - checkout
       - setup_remote_docker:
+          version: 17.03.0-ce
```
rowlando: Why downgrade Docker version?

sldblog: This is the latest known version which the registry works with. Once this is merged, we could explore upgrading it one by one.

sldblog:
🙇 🎉

PR #734, opened 2018-08-20
LGA-112: Add disk size to healthchecks

sldblog:

Extends the healthcheck endpoint with a disk size check.

It reports on the available and total disk space in megabytes and percent. It will **fail** the check when the available disk space gets at or under 2.0%.

💚 Test coverage has been added for `/healthcheck.json` and the two checks.

The response code of `/healthcheck.json` now changes to `503 Service Unavailable` if any of the checks return falsey (`False` or `None`).

`requests` and `urllib3` have been upgraded to their latest versions. I wanted to update `requests` only but `urllib3` is required both by that and our application so they had to be upgraded together.

### DIFF
--- a/cla_public/apps/base/healthchecks.py
+++ b/cla_public/apps/base/healthchecks.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+
+import requests
+from flask import current_app
+
+
+def check_backend_api():
+    backend_healthcheck_url = '%s/%s' % (
+        current_app.config['BACKEND_BASE_URI'], 'status/healthcheck.json')
+    status = False
+    response_content = None
+
+    try:
+        backend_healthcheck_response = requests.get(backend_healthcheck_url)
+        status = True if backend_healthcheck_response.ok else backend_healthcheck_response.error
+        response_content = backend_healthcheck_response.json()
+    except requests.exceptions.RequestException as e:
+        status = False
+        response_content = e.__class__.__name__
+
+    return {
+        'status': status,
+        'url': backend_healthcheck_url,
+        'response': response_content
+    }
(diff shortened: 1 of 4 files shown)


cla_public/apps/base/healthchecks.py
```diff
@@ -0,0 +1,43 @@
...
+def check_disk():
+    stat = os.statvfs(os.getcwd())
+    available_mb = (stat.f_bavail * stat.f_frsize) / (1024.0 ** 2)
+    total_mb = (stat.f_blocks * stat.f_frsize) / (1024.0 ** 2)
+
+    available_percent = available_mb / total_mb * 100
+    status = True if available_percent > 2.0 else False
+
+    return {
+        'status': status,
```
rowlando: I've run this on my local machine and these are the results:

```
{
  "Backend API test": {
    "response": "ConnectionError", 
    "status": false, 
    "url": "http://127.0.0.1:8000/status/healthcheck.json"
  }, 
  "disk": {
    "available_mb": 325536.171875, 
    "available_percent": 68.26059495331863, 
    "status": true, 
    "total_mb": 476902.04296875
  }
}
```

The `status: true` doesn't read well to me. I think something like: 

- `healthy: true` or `healthy: false`
- `status: healthy` or `status: unhealthy`

sldblog: Yeah, it's horrible. I followed the existing schema at the time.
Now that we know that it's only the HTTP response code that matters, we can change it.

rowlando: Is it worth `2.0` be a constant to make it more obvious that it's a threshold?

sldblog: ♻️

sldblog: > Is it worth `2.0` be a constant to make it more obvious that it's a threshold?

I feel in this current state it's readable enough to keep it as-is:

```
    status = HEALTHY if available_percent > 2.0 else UNHEALTHY
```

What do you think?

sldblog:
🎩 🙇

PR #747, opened 2018-08-28
Push to both DSD and ECR Docker registries

sldblog:

Push the built Docker image to both DSD and ECR Docker registries.

Unfortunately, using the DSD registry forces us to lock everything to Docker 17.03 where we want to use those images, which is proving to be a blocker.

In order to remove that constraint, we chose to push the built image into both Docker registries.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,13 +14,16 @@ jobs:
           version: 17.03.0-ce
           docker_layer_caching: true
       - run:
-          name: Login to DSD and ECR Docker registries
+          name: Login to the DSD Docker registry
+          command: |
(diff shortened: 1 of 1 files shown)


rowlando:
There also seems to be a problem with the build:

`Error parsing reference: "registry.service.dsd.io/:502bb4a8512bae0f0cc4c91807cd3110ac0ee060" is not a valid repository/tag`

sldblog:
@rowlando Yes, I made a mistake. It's fixed now

sldblog:
> Approved but wonder if it's worth separating the breaking up the login step into 2 steps, one for both registries, so it's clearer that the dependencies belong to ECR?

@rowlando Good idea! Done.

PR #753, opened 2018-09-11
Optimise Dockerfile

sldblog:

Optimises the `Dockerfile` by

- Installing node.js via https://deb.nodesource.com/setup_8.x.
- Copying minimal files required to install dependencies to avoid unnecessary reinstalling of python or npm packages any time a file changes.
- Renaming `ADD` to `COPY` as it is discouraged to use `ADD` for copying files.
- Moving EXPOSE and CMD to the end of the file as that is where they are conventionally located.
- Grouping commands related to `nginx` together.
- Grouping commands related to `uwsgi` together.
- Ordering packages alphabetically on separate lines as it makes it more readable and easier to find what is installed.
- Adding `LABEL` instructions as per http://docs.projectatomic.io/container-best-practices/#_recommended_labels_for_your_project.

**`--label` arguments have been added to the `docker build`**
They are viewable with

```
docker inspect --format='{{ json .ContainerConfig.Labels }} <tag>
```

Labels can be queried for running containers too:

```
docker ps --format "table {{.ID}}\t{{.Labels}}"
```

Ultimately, this convention is useful to find out the build information for the Docker image without mutating the image contents.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,10 @@
 # Pull base image.
 FROM phusion/baseimage:0.9.22
 
-# Set correct environment variables.
+LABEL name="Check If You Can Get Legal Aid (cla_public)" \
+      maintainer="LAA Get Access <laa-get-access@digital.justice.gov.uk>" \
+      version="1.0"
+
 ENV HOME /root
 # Use baseimage-docker's init process.
 CMD ["/sbin/my_init"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,8 +11,6 @@ LABEL name="Check If You Can Get Legal Aid (cla_public)" \
       version="1.0"
 
 ENV HOME /root
-# Use baseimage-docker's init process.
-CMD ["/sbin/my_init"]
 
 # Dependencies
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && \
@@ -46,10 +44,6 @@ COPY ./docker/nginx.conf /etc/nginx/nginx.conf
 #VOLUME ["/data", "/var/log/nginx", "/var/log/wsgi", "/var/log/cla_public"]
 RUN mkdir -p /var/log/nginx/cla_public
 
-# Expose ports.
-EXPOSE 80
-
-# APP_HOME
 ENV APP_HOME /home/app/flask
 WORKDIR /home/app/flask
 
@@ -68,3 +62,6 @@ COPY . .
 # Compile frontend assets and translations
 RUN ./node_modules/.bin/gulp build && \
     pybabel compile -d cla_public/translations
+
+EXPOSE 80
+CMD ["/sbin/my_init"]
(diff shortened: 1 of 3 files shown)


sldblog:
It looks like a `no-cache` build is about 10-20 seconds faster with these changes.
There should be significant speedup for cached builds.

Dockerfile
```diff
@@ -34,13 +34,10 @@ RUN pip install GitPython uwsgi
...
-
-# install service files for runit
-ADD ./docker/nginx.service /etc/service/nginx/run
+COPY ./docker/cla_public.ini /etc/wsgi/conf.d/cla_public.ini
+COPY ./docker/uwsgi.service /etc/service/uwsgi/run
+COPY ./docker/nginx.service /etc/service/nginx/run
+COPY ./docker/nginx.conf /etc/nginx/nginx.conf
 
 # Define mountable directories.
 #VOLUME ["/data", "/var/log/nginx", "/var/log/wsgi", "/var/log/cla_public"]
```
rowlando: While you're at it, should we delete this?

sldblog:
@rowlando Good points. Made further cleanups that group related instructions together. Please re-review 🙏

Dockerfile
```diff
@@ -6,70 +6,71 @@
...
 
-# Define mountable directories.
-#VOLUME ["/data", "/var/log/nginx", "/var/log/wsgi", "/var/log/cla_public"]
-RUN mkdir -p /var/log/nginx/cla_public
-
-# Expose ports.
-EXPOSE 80
-
-# APP_HOME
 ENV APP_HOME /home/app/flask
```
rowlando: Is this used anywhere?

sldblog: I am not sure. It sounds like it has a potential "dig through code" effect and as it does not add overhead to the Docker build I did not investigate. Could we defer it to a follow-up pull request?

Dockerfile
```diff
@@ -6,70 +6,71 @@
 # Pull base image.
 FROM phusion/baseimage:0.9.22
 
-# Set correct environment variables.
+LABEL name="Check If You Can Get Legal Aid (cla_public)" \
+      maintainer="LAA Get Access <laa-get-access@digital.justice.gov.uk>" \
+      version="1.0"
+
 ENV HOME /root
```
rowlando: Is this used anywhere?

sldblog: I think I would categorise this on the same level as `APP_HOME`. This might be a convention enforced by the base image and as the intent of this pull request is to optimise the Dockerfile, I glanced over it :) Could we defer it to another pull, please?

Dockerfile
rowlando: Sure

Dockerfile
rowlando: 👍

sldblog:
🎩 🙇

PR #758, opened 2018-09-14
set locale with query param

rowlando:

The application now reacts to a query parameter called `locale`. The locale can be set to `cy_GB` or `en_GB`.

This change was put in so that external sites, e.g. gov.uk, where the locale had already been set, could carry through to our application, resulting in a better user experience for the end user.

The query parameter can be on any URL in the application. It reuses the code which allowed the user to switch locale within our application. 

I have learnt that the application is made up of many [Flask Blueprints](http://flask.pocoo.org/docs/1.0/blueprints/). I hook into the Blueprint lifecycle event [before_app_request](http://flask.pocoo.org/docs/1.0/api/#flask.Blueprint.before_app_request).

I considered using middleware but this seemed simpler.


cla_public/libs/utils.py
```diff
@@ -28,6 +28,28 @@ def get_locale():
         language_keys
     ) or 'en'
 
+def change_locale(locale):
+    """
+    Set locale cookie
+    """
+    if locale[:2] not in [l for l, d in current_app.config['LANGUAGES']]:
+        abort(404)
```
sldblog: 💭1 - Would returning a 404 be correct if `locale` is a query parameter, or should we ignore an unrecognised locale instead?

💭2 - On that train of thought, should we replace the `/locale/<locale>` path with query parameters across the whole site for a consistent approach?

cla_public/apps/base/__init__.py
```diff
@@ -1,5 +1,12 @@
 "Base app"
 from flask import Blueprint
-
+from flask import request
+from cla_public.libs.utils import change_locale
```
sldblog: 💭 Could we move the locale-specific functions into a `cla_public.libs.locale` namespace instead?

cla_public/apps/base/__init__.py
```diff
@@ -1,5 +1,12 @@
...
 
 base = Blueprint('base', __name__)
+
+@base.before_app_request
+def before_app_request():
+  locale = request.args.get('locale')
+  if locale is not None:
+    return change_locale(locale)
```
sldblog: 👏

cla_public/apps/base/views.py
```diff
@@ -193,33 +193,10 @@ def get_started():
...
 
-
-def is_safe_url(url):
-    ref_url = urlparse(request.host_url)
-    test_url = urlparse(urljoin(request.host_url, url))
-    return test_url.scheme in ('http', 'https') and \
-        ref_url.netloc == test_url.netloc
-
-
-def next_url():
```
sldblog: There is another `next_url` defined in another view:
https://github.com/ministryofjustice/cla_public/blob/06ceb9dcc10041e040c75ac854d098f878281a9d/cla_public/libs/views.py#L204-L207

Moving this to `utils` could cause some (unlikely) unintended clashes with imports.

What do you think about renaming it to something more specific?

cla_public/apps/base/__init__.py
```diff
@@ -1,5 +1,16 @@
...
+
+@base.before_app_request
+def detect_user_locale():
+    locale = request.args.get('locale')
+    
+    if locale is not None:
+        @after_this_request
+        def remember_locale(response):
+            expires = datetime.datetime.now() + datetime.timedelta(days=30)
+            response.set_cookie('locale', locale, expires=expires)
```
sldblog: What do you think about removing the current path to set the locale, too? It seems with this changeset we have a duplication of code on setting this cookie.

rowlando: Will do in a separate pull request.

cla_public/apps/base/tests/test_locale.py
```diff
@@ -0,0 +1,30 @@
...
+  def setUp(self):
+    self.app = create_app('config/testing.py')
+    ctx = self.app.test_request_context()
+    ctx.push()
+    self.client = self.app.test_client()
+
+  def test_locale_cookie_is_set(self):
+    with self.app.test_client() as client:
+      response = client.get('/?locale=en_GB')
+      self.check_cookie(response, 'locale', 'en_GB')
```
sldblog: It feels we should also test the redirection here. What do you think?

cla_public/apps/base/__init__.py
```diff
@@ -1,5 +1,17 @@
...
+@base.before_app_request
+def detect_user_locale():
+    locale = request.args.get('locale')
+    
+    if locale is not None:
+        @after_this_request
+        def remember_locale(response):
+            expires = datetime.datetime.now() + datetime.timedelta(days=30)
+            response = redirect(request.path)
+            response.set_cookie('locale', locale, expires=expires)
```
sldblog: 💭 We set the cookie without validating the input argument. As a follow-up, should we restrict the allowed inputs to the two known supported locales?

cla_public/apps/base/tests/test_locale.py
rowlando: Sure.

cla_public/apps/base/__init__.py
rowlando: I'm not sure if it's worth it. The current behaviour is that if the locale isn't recognised by babel, it defaults to 'en'.
Do you think there's a risk if we don't validate the input? If anything breaks, it would be because someone is playing with it.

sldblog: Hmm, I think the "path" selector sanitises and that's why frontend relies on a binary check:
https://github.com/ministryofjustice/cla_public/blob/0eef40a8ce87df94451ac5f95f4dcb6d210d6b62/cla_public/apps/base/views.py#L215-L216

I think it's a good idea to sanitize all user inputs. Cookies have a maximum size and people can be very clever to cause grief :)

rowlando: Will add this to a tech debt ticket as discussed.

cla_public/apps/base/tests/test_locale.py
rowlando: Will add this to a tech debt ticket as discussed.

cla_public/apps/base/__init__.py
rowlando: LGA-192

PR #759, opened 2018-09-17
LGA-192 Add redirect to gov.uk start page

rowlando:

This pull request redirects the route `/` to https://www.gov.uk/check-legal-aid, the GOV.UK start page for Civil Legal Advice.

It is not possible to change the link in the header directly because it is buried inside `base.jinja` template, which lives inside the `jinja-moj-template` package.

None.

### DIFF
--- a/cla_public/config/common.py
+++ b/cla_public/config/common.py
@@ -11,6 +11,8 @@ TESTING = False
 
 CLEAR_SESSION = True
 
+GOV_UK_START_PAGE = 'https://www.gov.uk/check-legal-aid'
+
 BACKEND_BASE_URI = os.environ.get('BACKEND_BASE_URI', 'http://127.0.0.1:8000')  # For healthcheck.json requests
 
 # Disable eligibility check and allow users to request a callback only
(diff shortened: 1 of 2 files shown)


sldblog:
_nitpick_ Could you link the JIRA ticket number in the pull request title, please?

cla_public/config/common.py
```diff
@@ -11,6 +11,8 @@
 
 CLEAR_SESSION = True
 
+GOV_UK_START_PAGE = 'https://www.gov.uk/check-legal-aid'
```
sldblog: This will sound extremely :neckbeard:y and we don't need to solve this now.

Do we need to redirect to https://www.gov.uk/gwirio-os-ydych-yn-gymwys-i-gael-cymorth-cyfreithiol (Welsh) if our locale is set to Welsh?

rowlando: I thought about that but then thought it might be more trouble than it's worth right now. It introduces another coupling between the two systems.

If we wanted to do this well, we'd want to be alerted if either of these link changes. Or check if there's an API where we can get a list of the start pages for an application.

rowlando:
> @rowlando It will be inaccessible for the public and might be misleading to anyone searching GitHub or this repository for words related to it. Should we delete it anyway and figure out a way to demo if we need to?

Fine by me.

PR #761, opened 2018-09-18
Notify Slack channel on deployments

sldblog:

Adds a notification to our Slack channel about deployments. For now, only _staging_ deploys are supported.

screen shot 2018-09-18 at 11 11 19

#### Moved all lengthy CI scripts to separate files in the `.circleci` folder

Docker builds and deployments share a lot of environment variables. One way would be to redefine all the variables. Another would be to reuse the definitions.

We took the approach of reusing the definitions from a shared shell script which is `source`d into all CI scripts.

#### Extend CI scripts with explicit messages

Moving the scripts to files has a side-effect of the deployment script showing only this output:

```
$ #!/bin/bash -eo pipefail
  .circleci/deploy_to_staging

deployment.extensions "laa-cla-public" configured
service "laa-cla-public" unchanged
ingress.extensions "laa-cla-public" unchanged
```

This information is not useful enough to see _what_ was deployed. We extended the deploy script with an explicit message about what is being deployed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,26 +40,7 @@ jobs:
           command: docker run --rm --tty --interactive application:$CIRCLE_SHA1 date '+%Z' | egrep "(GMT|BST)"
       - run:
           name: Tag and push Docker images
-          command: |
-            built_tag="application:$CIRCLE_SHA1"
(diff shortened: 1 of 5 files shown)


sldblog:
🎩 🙇 Thank you!

PR #760, opened 2018-09-17
LGA-196: Log to stdout

rowlando:
## What does this pull request do?

This pull request configures "nginx" and "uwsgi" to log to stdout in JSON format.

## Any other changes that would benefit highlighting?

Note that "nginx" does not log the timestamp to millisecond resolution. Instead of `$time_iso8601`, we could choose to log `$msec`, which is defined as:

>time in seconds with a milliseconds resolution at the time of the log write 

Below is an example of the log entries for nginx, uwsgi and the application. 

### Nginx log entry

```
{
  "@timestamp": "2018-09-19T11:33:01+01:00",
  "@fields": {
    "process_name": "nginx",
    "pid": 27,
    "remote_addr": "192.168.144.1",
    "remote_user": "-",
    "body_bytes_sent": 17,
    "request_time": 0.003,
    "request_uri": "/session_keep_alive",
    "status": 200,
    "request": "GET /session_keep_alive HTTP/1.1",
    "request_method": "GET",
    "http_referrer": "http://localhost:5000/session-expired",
    "http_user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36",
    "bytes_sent": 453,
    "gzip_ratio": "-",
    "http_host": "localhost",
    "sent_http_location": "-",
    "server_name": "~^(.+)$",
    "server_port": "80",
    "upstream_addr": "unix:////tmp/public.sock",
    "upstream_response_length": "17",
    "upstream_response_time": "0.010",
    "upstream_status": "200"
  }
}
```

### uwsgi log entry

```
{
  "process_name": "uwsgi",
  "timestamp_msec": 1537353181294496,
  "method": "GET",
  "uri": "/session_keep_alive",
  "proto": "HTTP/1.1",
  "status": 200,
  "referer": "http://localhost:5000/session-expired",
  "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36",
  "remote_addr": "192.168.144.1",
  "http_host": "localhost:5000",
  "pid": 35,
  "worker_id": 1,
  "core": 0,
  "async_switches": 1,
  "io_errors": 0,
  "rq_size": 0,
  "rs_time_ms": 2,
  "rs_size": 288,
  "rs_header_size": 271,
[...]

PR #762, opened 2018-09-19
LGA-196 Add new line after each uWSGI log message

rowlando:

This pull request adds a new line after each uWSGI log message. It also 

We also have switched core log encoder from `json` to `format`. See https://uwsgi-docs.readthedocs.io/en/latest/LogEncoders.html#core-encoders for info.

We don't need to escape each variable.

*Edit 25/9/2018* We have switched back to the default encoder and we format the message as JSON line-by-line.


sldblog:
I'm curious, does this provide valid JSON output? If `format {"${msgnl}"}` does not escape variables for JSON, I'm not following how the quotes inside the message will be represented as the format itself is using quotes?

rowlando:
> I'm curious, does this provide valid JSON output? If `format {"${msgnl}"}` does not escape variables for JSON, I'm not following how the quotes inside the message will be represented as the format itself is using quotes?

When would quotes be used inside the message? I guess `user_agent` would be the only likely candidate.

Happy to change it back to be on the safe side.

rowlando:
https://github.com/ministryofjustice/cla_public/pull/760 shows what the various logs output.

I'm wondering if we should remove the request variable from the uwsgi log entry? E.g. `method`, `referer`, `user_agent`, etc..

```
  "process_name": "uwsgi",
  "timestamp_msec": 1537353181294496,
  "method": "GET",
  "uri": "/session_keep_alive",
  "proto": "HTTP/1.1",
  "status": 200,
  "referer": "http://localhost:5000/session-expired",
  "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36",
  "remote_addr": "192.168.144.1",
  "http_host": "localhost:5000",
  "pid": 35,
  "worker_id": 1,
  "core": 0,
  "async_switches": 1,
  "io_errors": 0,
  "rq_size": 0,
  "rs_time_ms": 2,
  "rs_size": 288,
  "rs_header_size": 271,
  "rs_header_count": 3
}
```

sldblog:
In the current version, there are two things I noticed:

screen shot 2018-09-25 at 11 36 57

- The `wsgi` message ends with an espaced newline: `\n"}`
- The `wsgi` message contents are escaped in a string, which means the content is invalid JSON (red is the current, green is the valid variant):
```diff
-"timestamp_msec": 1537871658904646, "\"method\": \"GET\", ...
+"timestamp_msec": 1537871658904646, "method": "GET", ...
```

sldblog:
It seems formatting the `logformat` as a JSON message and dropping the `log-req-encoder` works:
```diff
-log-req-encoder=json {"process_name": "uwsgi", "timestamp_msec": ${micros}, "${msgnl}"}
-logformat="method": "%(method)", "uri": "%(uri)", "proto": "%(proto)", "status": %(status), "referer": "%(referer)", "user_agent": "%(uagent)", "remote_addr": "%(addr)", "http_host": "%(host)", "pid": %(pid), "worker_id": %(wid), "core": %(core), "async_switches": %(switches), "io_errors": %(ioerr), "rq_size": %(cl), "rs_time_ms": %(msecs), "rs_size": %(size), "rs_header_size": %(hsize), "rs_header_count": %(headers)
+logformat={"process_name": "uwsgi", "timestamp_msec": %(tmsecs), "method": "%(method)", "uri": "%(uri)", "proto": "%(proto)", "status": %(status), "referer": "%(referer)", "user_agent": "%(uagent)", "remote_addr": "%(addr)", "http_host": "%(host)", "pid": %(pid), "worker_id": %(wid), "core": %(core), "async_switches": %(switches), "io_errors": %(ioerr), "rq_size": %(cl), "rs_time_ms": %(msecs), "rs_size": %(size), "rs_header_size": %(hsize), "rs_header_count": %(headers)}
```

PR #751, opened 2018-09-10
Add staging-deploy to CircleCI

AbdirizakObsiye2018:
## What does this pull request do?

Adds staging-deploy to CircleCI behind a manual approval step.

This has been mostly copied from the [laa-fee-calculator](https://github.com/ministryofjustice/laa-fee-calculator) project.

📝 We are using `kubectl` commands to deploy into staging environment before `helm` because it's quicker to implement. The intention is to use Helm in the future.

Pairing with: @sldblog

## Implementation notes

To make `kubectl config use-context staging` work, we had to create a `circleci` `ServiceAccount` in https://github.com/ministryofjustice/cloud-platform-environments/pull/140 and create the following environment variables in CircleCI:

- `KUBE_ENV_STAGING_CACERT`
- `KUBE_ENV_STAGING_NAME`
- `KUBE_ENV_STAGING_NAMESPACE`
- `KUBE_ENV_STAGING_TOKEN`

This is documented in [cloud-platform-user-docs](https://github.com/ministryofjustice/cloud-platform-user-docs/blob/648a435250eb11bb07b64b280249ca05db156943/02-deploying-an-app/004-use-circleci-to-upgrade-app.md#add-variables-to-circleci).

Using the `setup-kube-auth` script bundled in the `cloud-platforms/tools:circleci` image sets up the "staging" context based on these environment variables:

screen shot 2018-09-13 at 12 48 23

PR #763, opened 2018-09-21
Add production-deploy to CircleCI

rowlando:

This pull requests builds on top of https://github.com/ministryofjustice/cla_public/pull/751. 

Circle CI has been configured to only allow `master` branch to run the `production_deploy` job.

The readme has been updates with instructions on how to deploy to Kubernetes production environment.

https://github.com/ministryofjustice/cla_public/blob/kubes-prod-circleci/README.md

screen shot 2018-09-24 at 12 34 27

Intentionally left blank.


.circleci/config.yml
```diff
@@ -90,14 +94,30 @@ jobs:
             kubectl config use-context staging
       - deploy:
           name: Deploy laa-cla-public to staging
-          command: .circleci/deploy_to_staging
-      - run:
+          command: .circleci/deploy_to_kubernetes staging
+      - deploy:
```
sldblog: 👏

.circleci/deploy_to_kubernetes
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh -eu
+
+ROOT=$(dirname "$0")
+ENVIRONMENT="$1"
+ENVIRONMENT_DIR="$ROOT/../kubernetes_deploy/$ENVIRONMENT"
+
+if ! [ -d $ENVIRONMENT_DIR ] ; then
+  echo "$ENVIRONMENT_DIR not found"
```
sldblog: Would we enter this branch if `$1` is empty (missing)? It feels like `$ROOT/../kubernetes_deploy/` would pass this test.

rowlando: Note the `-u` in `#!/bin/sh -eu`.

>“set -u” makes the script exit if a referenced variable is not declared. Pretty useful since an undeclared variable can be mistaken for an empty one.

So in this case, `$1` isn't declared.

sldblog: TIL 😍 

💭 In this case, what do you think about also printing a usage line when the argument is missing? Might be easier for the users.

PR #765, opened 2018-10-02
LGA-284 Configure Kubernetes to use HTTP liveness-probe

rowlando:

This pull requests configures Kubernetes to use liveness-probe to determine the health of the application.

The liveness probe will return a success if the application end point `ping.json` responds with a 200. If it doesn't, Kubernetes will attempt to restart the container. 

I have also set replicas to 2 for `staging` and 3 for `production`. This is [Kerin's comment](https://mojdt.slack.com/archives/CACTAK5FG/p1538486123000100) on why 3 is a good number:

>Yep default is one - personally I’d recommend 3, so you can still tolerate a failure during a deployment
>That also matches the number of AZs in AWS Ireland, and your containers will get spread across them evenly

By having pod replicas, Kubernetes will not terminate a running pod until deployed containers are running and ready, as deemed by the readiness check. This results in no downtime between releases. This image shows the lifecycle of the pods:

screen shot 2018-10-02 at 16 36 43

I have upped the seconds in which the readiness check first takes place so it safely takes longer than the time the app usually takes to start up. At the moment, the app takes a few seconds to start.

We can set `staging` replicas back to one. It's set to 2 at the moment to see it working before this pull request is merged with master.

### DIFF
--- a/kubernetes_deploy/production/deployment.yml
+++ b/kubernetes_deploy/production/deployment.yml
@@ -15,8 +15,16 @@ spec:
           httpGet:
             path: /ping.json
             port: 80
-          initialDelaySeconds: 3
-          periodSeconds: 3
+          initialDelaySeconds: 5
+          timeoutSeconds: 1
+          periodSeconds: 10
+        livenessProbe:
+          httpGet:
+            path: /ping.json
+            port: 80
+          initialDelaySeconds: 10
+          timeoutSeconds: 1
+          periodSeconds: 10
         ports:
         - containerPort: 80
           name: http
--- a/kubernetes_deploy/production/deployment.yml
+++ b/kubernetes_deploy/production/deployment.yml
@@ -8,6 +8,7 @@ spec:
       labels:
         app: laa-cla-public-app
     spec:
+      replicas: 3
       containers:
       - image: 926803513772.dkr.ecr.eu-west-1.amazonaws.com/laa-get-access/laa-cla-public:master
         name: app
--- a/kubernetes_deploy/production/deployment.yml
+++ b/kubernetes_deploy/production/deployment.yml
@@ -3,12 +3,12 @@ kind: Deployment
 metadata:
   name: laa-cla-public
 spec:
+  replicas: 3
   template:
     metadata:
       labels:
         app: laa-cla-public-app
     spec:
-      replicas: 3
       containers:
       - image: 926803513772.dkr.ecr.eu-west-1.amazonaws.com/laa-get-access/laa-cla-public:master
         name: app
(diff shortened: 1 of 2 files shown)


kubernetes_deploy/production/deployment.yml
```diff
@@ -15,8 +16,16 @@ spec:
           httpGet:
             path: /ping.json
```
sldblog: *not a blocker*

I wonder if there anywhere where we could use `/healthcheck.json`. When a liveness probe fails, the container gets restarted. When a readiness probe fails, the container gets pulled out of traffic.

Is there any semantic where any of the checks in the health check necessitate the above behaviours?

kubernetes_deploy/production/deployment.yml
```diff
@@ -15,8 +16,16 @@ spec:
...
             port: 80
-          initialDelaySeconds: 3
-          periodSeconds: 3
+          initialDelaySeconds: 5
+          timeoutSeconds: 1
+          periodSeconds: 10
```
sldblog: I'm curious, could you expand on why these timeout and period durations were chosen, please? (I found the initial delay mentioned in the description.)

kubernetes_deploy/production/deployment.yml
```diff
@@ -3,6 +3,7 @@ kind: Deployment
 metadata:
   name: laa-cla-public
 spec:
+  replicas: 3
```
sldblog: 👏 Thank you for the explanation in the description!

kubernetes_deploy/production/deployment.yml
rowlando: I'm not sure. This is a sample response of `/healthcheck.json`:

```json
{
  "Backend API test": {
    "response": {
      "*": {
        "status": true
      }, 
      "database": {
        "status": true
      }
    }, 
    "status": "healthy", 
    "url": "https://fox.civillegaladvice.service.gov.uk/status/healthcheck.json"
  }, 
  "disk": {
    "available_mb": 76143.25, 
    "available_percent": 75.64691463630018, 
    "status": "healthy", 
    "total_mb": 100656.12109375
  }
}
```

Let's clarify what happens when a container's liveness or readiness probe fails:

>A Pod is considered ready when all of its Containers are ready. One use of this signal is to control which Pods are used as backends for Services. When a Pod is not ready, it is removed from Service load balancers.

Given this, would tightly coupling the application's health to the health of the Backend API be the correct behaviour?

If the Backend API is down, the healthcheck will fail for all replicas, taking the service offline. Is this correct? Or should the application remain up? I believe the UI handles this situation.

Are there any cases where an application health check could fail in one replica and not another?

Is the `disk` check useful in a Kubernetes world?

kubernetes_deploy/production/deployment.yml
rowlando: These are the Kubernetes defaults made explicit. See https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes

PR #766, opened 2018-10-30
LGA-206 "Use Kubernetes" section in README

rowlando:

This pull request adds a Use Kubernetes section to the readme. See https://github.com/ministryofjustice/cla_public/blob/document-kubernetes/README.md

### README is too long
Screenshot of previous readme

### README broken up into sections
Screenshot of readme

Soon, we should update the `deploy_to_kubernetes` script so both arguments are passed, much like FALA. But I think this can be done once we get to the point of DRYing the scripts up and adding them to the tools image.


README.md
```diff
@@ -115,7 +171,7 @@ where the `context` directory is set to the root of the cla_public directory.
 
 :tada: :shipit:
 
-#### Kubernetes Deploy
+#### Deploy to Kubernetes using Circle CI
```
sldblog: _nitpick_ I think the service name is written as "CircleCI" (one word): https://circleci.com/docs/

README.md
```diff
@@ -83,6 +114,31 @@ cla_public:
...
+
+You'll need to install and configure `kubectl` CLI tool to interact with Kubernetes. There are [instructions on kubectl configuration](https://ministryofjustice.github.io/cloud-platform-user-docs/01-getting-started/001-kubectl-config/) in the Cloud Platform User Guide.
+
+### Kubernetes namespaces and admin role
+`cla_public` has two namespaces (environments):
+
+- [laa-cla-public-staging](https://github.com/ministryofjustice/cloud-platform-environments/tree/master/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/laa-cla-public-staging)
+- [laa-cla-public-production](https://github.com/ministryofjustice/cloud-platform-environments/tree/master/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/laa-cla-public-production)
+
+To gain the `ClusterRole - admin` role, you must be a member of the GitHub team `laa-get-access` - [find out more about roles](https://ministryofjustice.github.io/cloud-platform-user-docs/01-getting-started/002-env-create/#01-rbacyaml).
```
sldblog: 🙋‍♂️ It might be a good thing to explain why `ClusterRole - admin` role is necessary. What does it let us do?

README.md
```diff
@@ -83,6 +114,31 @@ cla_public:
...
+
+### Authenticating with AWS ECR repository
```
sldblog: 🙋‍♂️ What do you think about saying "Authenticating with the Docker registry"? AWS ECR is the implementation of a registry, while people might search for "Docker" if they're interested in that.

README.md
```diff
@@ -83,6 +114,31 @@ cla_public:
...
+Docker images are stored in AWS ECR. To authenticate with the `cla_public` repository, fetch the credentials by typing the following:
+
+```
+kubectl --namespace laa-cla-public get secrets -o yaml
```
sldblog: ❓ Would we need to use either `laa-cla-public-staging` or `laa-cla-public-production` here?

README.md
```diff
@@ -83,6 +114,31 @@ cla_public:
...
+```
+
+This command will return the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. You can find out more by reading [Authenticating with the repository](https://ministryofjustice.github.io/cloud-platform-user-docs/02-deploying-an-app/001-app-deploy/#authenticating-with-the-repository)
```
sldblog: 🙋‍♂️ I think it's worth stressing that these values are encoded by saying "This command will return the **encoded** `AWS_ACCESS_KEY_ID` ...", etc. What do you think?

README.md
```diff
@@ -1,7 +1,41 @@
 # CLA Public
 
+## Important notices
+
+- `master` branch is the default branch
```
sldblog: `master` as the default branch seems to be the standard practice across GitHub repositories. Should we mention it?

README.md
```diff
@@ -1,7 +1,41 @@
...
+
+## Table of contents
```
sldblog: We will probably forget to update the table of contents on any subsequent update in the contents. (I would.) It is helpful, but perhaps we should consider moving some of these to a `docs/` folder or wiki?

README.md
rowlando: `laa-cla-public-staging` - good spot.

README.md
rowlando: We could use something like https://ecotrust-canada.github.io/markdown-toc/ to generate manually.

If developers find a TOC helpful, we could think about automating this.

I think a new developer eyes will be handy to compare this README with READMEs in our other repositories.

sldblog: Sorry, what I meant was that the pattern of _needing_ a TOC means the readme is too large.

What do you think about separating each top-level section (dependencies, installation, development, testing, etc.) into separate files:

```
README.md
docs/
  dependencies.md
  installation.md
  ... etc
```

And adding links to them, ie. `[Dependencies](docs/dependencies.md)`?

This way we would have a structured readme but the cost of maintaining the structure is low and intuitive, without the need of external tools.

sldblog:
📝 GitHub status does not seem to be matching CircleCI

screen shot 2018-11-02 at 12 43 28

README.md
rowlando: Good idea!

PR #767, opened 2018-11-30
LGA-380 Custom working hours for 2018-12

farrepa:

Set custom working hours to match Operator Service's reduced hours on Christmas Eve and New Year's Eve 2018.

See accompanying changes in related repos:

- [cla_backend#488](
https://github.com/ministryofjustice/cla_backend/pull/488)
- [cla_common/call_centre_availability/__init__.py#L173](https://github.com/ministryofjustice/cla_common/blob/0.2.7/cla_common/call_centre_availability/__init__.py#L173)

### DIFF
--- a/cla_public/config/common.py
+++ b/cla_public/config/common.py
@@ -106,8 +106,8 @@ FORM_CONFIG_TRANSLATIONS = {l: config_path(l) for l, label in LANGUAGES}
 OPERATOR_HOURS = {
     'weekday': (datetime.time(9, 0), datetime.time(20, 0)),
     'saturday': (datetime.time(9, 0), datetime.time(12, 30)),
-    '2014-12-24': (datetime.time(9, 0), datetime.time(18, 30)),
-    '2014-12-31': (datetime.time(9, 0), datetime.time(18, 30)),
+    '2018-12-24': (datetime.time(9, 0), datetime.time(17, 30)),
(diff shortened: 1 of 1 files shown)


sldblog:
> Set custom working hours for 2018-02

_nitpick_ `2018-12`

Could we also include the context (JIRA ticket number, or explaining) please? I find it useful to not only see the facts but also understand the drive behind the changes when I am investigating a diff in the future.

I also find cross-referencing useful in case there are similar work to do in the future.

What do you think about the following?

> Set custom working hours for 24 December (Christmas Eve) and 31 December (New Year's Eve) for 2018 as the Operator Service is planning to do reduced hours on these days.
> 
> See also:
> - https://github.com/ministryofjustice/cla_backend/pull/488
> - https://github.com/ministryofjustice/cla_common/blob/0.2.7/cla_common/call_centre_availability/__init__.py#L173

farrepa:
Sure. Depends on the balance you'd like to strike between referencing external closed systems and partially transcribing tickets here.

If you're comfortable prefixing ticket numbers to branch names that's a much easier mechanism to reference the rationale and backstory for commits.

sldblog:
I personally prefer to explain the reasons independently.

I think we require the linking, but I believe the changes have to make sense on their own, without digging into change management software.

PR #779, opened 2019-01-14
LGA-350: Add warning message about split deployments

sldblog:

We need to remember to deploy to both environments while the split between template-deploy and Kubernetes is running.

This change mentions this in the deployment readme.

None.


sldblog:
:tophat:

PR #792, opened 2019-01-17
LGA-457: Add Google Analytics ID to Kubernetes

sldblog:

This variable missing does not cause any errors on booting up the application, so we did not detect it before.

As we now have 10% traffic over to Kubernetes, quickly adding `GA_ID` environment variable to the production Kubernetes configuration.

None.


sldblog:
🎩

PR #796, opened 2019-01-22
LGA-450 Fix map with previously assigned lambda and add simple test

farrepa:

- Fix `category_id_to_name` to expect a single unexpanded category.
- Add a simple covering test.

Intentionally left blank.

### DIFF
--- a/cla_public/libs/tests/test_utils.py
+++ b/cla_public/libs/tests/test_utils.py
@@ -1,0 +1,10 @@
+import unittest
+
+from cla_public.libs.utils import category_id_to_name
+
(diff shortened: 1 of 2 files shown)


sldblog:
Sharing the context to help understanding in the future.

### Why was this change necessary?

It looks like after merging bea3e9b, we started receiving 5xx errors: https://sentry.service.dsd.io/mojds/prod-public/issues/34416/ `TypeError: selected_name() takes exactly 3 arguments (1 given)`.

screen shot 2019-01-22 at 10 21 41

PR #797, opened 2019-01-22
Annotate deployments with change-cause

sldblog:

Automatically annotate the deployments caused by CircleCI so we can see the rollout history:

```
$ kubectl rollout history deployment/laa-cla-public --namespace=laa-cla-public-production
deployment.extensions/laa-cla-public
REVISION  CHANGE-CAUSE
1         <none>
2         <none>
3         https://circleci.com/gh/ministryofjustice/cla_public/960
```

The intention is to easily identify why deployments happened and link back to the job that caused the rollout.

Reference: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#checking-rollout-history-of-a-deployment

Added `-o pipefail` to the shebang. If any of the `kubectl set` or `kubectl annotate` commands fail, we should fail the deployment.

Example: https://circleci.com/gh/ministryofjustice/cla_public/964 was green, despite the failed commands.


sldblog:
🎩

PR #799, opened 2019-01-23
LGA-351: Update Grafana dashboard

sldblog:

- Locks legend to 450 pixels wide, so the graph area is consistent in width.
- Shows "maximum" values in legends for CPU/memory and network.
- Smooths the graph displayed by taking data over 5 minutes instead of 1 minute.
- Defaults the dashboard to the last 24 hours instead of last 5 minutes.

**Before**
screen shot 2019-01-23 at 12 26 13

**After**
screen shot 2019-01-23 at 12 33 21

- Changes production dashboard UID (just in case it matters).


sldblog:
🎩

PR #801, opened 2019-01-24
LGA-481: Remove template-deploy Docker registry from build

sldblog:

We are 100% on Kubernetes! ~~(Well, not _yet_, but it will be applied today.)~~

We no longer need to deploy on template-deploy, so we no longer need to build for the Docker registry used by that environment.

This pull request removes the use of the old registry.

The change speeds up the Docker build job by about 2.5 minutes (7 minutes 50 seconds &rarr; 5 minutes 10 seconds).

## Dependencies

- [x] 🚨 ~~Do not merge until Kubernetes 100% split is applied, please.~~ 100% split is live.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,21 +12,14 @@ references:
 jobs:
   build:
     docker:
-      - image: docker:17.03-git
+      - image: docker:18.09-git
     environment:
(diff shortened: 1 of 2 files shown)


sldblog:
🎩

PR #802, opened 2019-01-24
LGA-481: Remove template-deploy from documentation

sldblog:

This pull request removes documentation about the template-deploy environment.

We have switched to Kubernetes (provided by the Cloud Platform team at MoJ Digital), which means we no longer need the previous environment.

None.


sldblog:
🎩

PR #807, opened 2019-02-08
LGA-497 - Stop application from booting up without essential configuration

said-moj:

Currently, it’s possible for the application to be deployed without having

- SMTP (confirmation e-mail) configuration
- Zendesk (feedback) configuration
- Sentry (exception monitoring) configuration

This leads to application silently failing to do any of the above tasks.

This PR prevents the application starting up without those configurations

It seems the debug toolbar extension is registered when DEBUG is True. Also having the `EXTENSIONS.append(DebugToolbarExtension())` in the local.py was causing app to fail complaining about the view method already being registered. I have commented this snippet out

The opening hours config was moved to a separate config file as this is required before the flask app has been created. The previous method was getting the open hours settings directly from commons.py instead of reading the current application config.

Added sensible default configuration for local and test environments.

### DIFF
--- a/cla_public/config/local.py.example
+++ b/cla_public/config/local.py.example
@@ -17,5 +17,3 @@ DEBUG = True
 SESSION_COOKIE_SECURE = False
 
 DEBUG_TB_INTERCEPT_REDIRECTS = False
-#EXTENSIONS.append(DebugToolbarExtension())
-
--- a/cla_public/config/local.py.example
+++ b/cla_public/config/local.py.example
@@ -1,6 +1,15 @@
 from flask_debugtoolbar import DebugToolbarExtension
 
-from .common import *
+import os
+os.environ.setdefault("SMTP_HOST", "")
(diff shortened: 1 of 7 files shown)


cla_public/config/local.py.example
```diff
@@ -1,18 +1,21 @@
...
+from cla_public.config.common import *
 
 SECRET_KEY = 'secret'
 
 DEBUG = True
 SESSION_COOKIE_SECURE = False
 
 DEBUG_TB_INTERCEPT_REDIRECTS = False
-EXTENSIONS.append(DebugToolbarExtension())
+#EXTENSIONS.append(DebugToolbarExtension())
```
sldblog: Can we remove this line instead of commenting, please?

docs/installation.md
```diff
@@ -24,6 +24,8 @@ Next, create the environment and start it up:
 
     npm install -g gulp
 
+    npm install && gulp
```
sldblog: 👏

PR #815, opened 2019-02-13
Allows forked pull requests to build Docker images

sldblog:

Allows forked pull requests to build Docker images without pushing.

The end goal is a workflow that runs all the tests but does not produce testable artefacts. I think this is a workable state for public contributions.

The `$CIRCLE_PR_NUMBER` environment variable is only populated on forked pull requests. (https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-project)

Using this behaviour we can split the control flow of the Docker build step to build and test the Docker image, but do not attempt to login and push.

Initially, I wanted to separate the jobs to (`docker-build` and `docker-push`) but then I realised making the image built in the first job available for the second job would not be trivial, so I lazied out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,10 +23,14 @@ jobs:
       - run:
           name: Login to the ECR Docker registry
           command: |
-            apk add --no-cache --no-progress py2-pip
-            pip install awscli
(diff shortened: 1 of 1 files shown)


sldblog:
Test test in #816 was successful.

sldblog:
🎩

PR #817, opened 2019-02-14
LGA-447 Remove Monday morning callback capping

farrepa:

> **NOTE: Do not merge until removal of callback capping has been confirmed with Agilisys.**

- Remove Monday morning call-back capping

- Retain and invert tests to ensure callback capping is removed
- Fix previous test so `tuesday` is actually a Tuesday(!)


cla_public/apps/contact/tests/test_availability.py
```diff
@@ -178,8 +175,8 @@ def test_day_time_choices(self):
             choices = field.day_time_choices
             # half day on saturday
             self.assertEqual(7, len(choices["20150214"]))
-            # can't book before 11am on monday because we're after hours friday
-            self.assertEqual(18, len(choices["20150216"]))
+            # can book before 11am on monday. Monday morning call back capping removed.
+            self.assertEqual(22, len(choices["20150216"]))
```
sldblog: Is it worth having a follow-up (once this has been live for a month) to remove the testing Mondays? It seems it can be replaced with testing for "weekdays".

PR #821, opened 2019-02-21
Local Kubernetes "one-node cluster" deployment

sldblog:

Recently, we had our content designer to set up the application locally, however for a non-developer, it seemed too involved for us.

This pull request adds the ability to build and run the application in a locally running one-node Kubernetes cluster. Solutions like `minikube` and Docker for (Mac|Windows) have made local Kubernetes deployments reasonably convenient.

We expect that this setup will make life easier for anyone who does not need to develop the codebase.

### Needs VPN to work with the case handler API

- The configuration currently is wired up to connect to staging backend services; there's no attempt to run the entire set of applications locally (although we could, it should probably belong in https://github.com/ministryofjustice/laa-cla-e2e-tests).
- The above means MoJ VPN has to be active. Empirically, the VPN has to be active before the pod starts up.

### Biggest differences compared to staging/production

- There is no _ingress_ resource defined because the service is exposed via `NodePort` to make it locally accessible on http://localhost:30000.
- The`imagePullPolicy` is set to `Never` because the setup assumes the image is available on the same Docker host -- otherwise, we would have to push the image to a registry.

### Other changes

- Stubbed "default" `image` entries in `deployment.yml`. This makes it clearer that they are set in the `deploy_to_kubernetes` script.
- We do not use secrets in the "local" configuration; instead, we set every third-party token to empty.
- **Bonus**: removed the "Jenkins" related application config.


sldblog:
📝 Tested with @joshlow. It worked 🎉

sldblog:
@farrepa ?

farrepa:
Looks good to me. Since it works locally for Josh, I shan’t run it locally (and bodge it to do so outside the VPN).

Out of curiosity, given that both Kubernetes and docker-compose ship with the Docker app, any particular reason for the more complex Kubernetes approach?

sldblog:
@farrepa Thank you!

> Out of curiosity, given that both Kubernetes and docker-compose ship with the Docker app, any particular reason for the more complex Kubernetes approach?

I was considering `docker-compose`, but since we already abstracted the use of `kubectl` to `deploy_to_kubernetes` script it seemed better to use the existing "tech" rather than to create another way of achieving a similar thing.

I favour the similarity: if someone wants to add a new runtime environment variable, they can do that in a single `kubernetes_deploy` folder.

sldblog:
🎩 🙇

PR #822, opened 2019-02-25
LGA-422 Set autocomplete and type attributes on contact_number input

farrepa:

Add `autocomplete="tel"` and `type="tel"` to `contact_number` input fields.

This is to fix "callback requests are being received with no phone number recorded against the case".

Intentionally left blank.

### DIFF
--- a/cla_public/templates/macros/subform.html
+++ b/cla_public/templates/macros/subform.html
@@ -20,7 +20,7 @@
         <li>{{ _('If you miss the call you’ll need to complete this form again.') }}</li>
       </ul>
     </div>
-    {{ Form.group(form.callback.contact_number, field_attrs={'class': 'm-large'}) }}
+    {{ Form.group(form.callback.contact_number, field_attrs={'class': 'm-large', 'autocomplete': 'tel', 'type': 'tel'}) }}
     {{ Form.group(form.callback.safe_to_contact) }}
     {{ time_picker(form.callback.time) }}
   {% endcall %}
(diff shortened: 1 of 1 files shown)


sldblog:
📝 Added a line to refer to the reason why these changes are made.

cla_public/templates/macros/subform.html
```diff
@@ -20,7 +20,7 @@
         <li>{{ _('If you miss the call you’ll need to complete this form again.') }}</li>
       </ul>
     </div>
-    {{ Form.group(form.callback.contact_number, field_attrs={'class': 'm-large'}) }}
+    {{ Form.group(form.callback.contact_number, field_attrs={'class': 'm-large', 'autocomplete': 'tel', 'type': 'tel'}) }}
```
sldblog: I'm not fluent in the art of `field_attrs` 😞 
Could you explain how will this solve the problem, please?

farrepa: It's a dict of extra attributes to add (or merge with existing attributes values) on the rendered input element.

The `contact_number` input element is used to request a phone number to call users back. It will now be rendered with an `autocomplete="tel"` attribute, as per the guidance at https://design-system.service.gov.uk/patterns/telephone-numbers/

Additionally we set `type="tel"` for the extra nicety of optimising keyboard entry on mobile browsers.

This is the [quick win agreed on](https://mojdt.slack.com/archives/CFUESB43G/p1550848531052500?thread_ts=1550837360.034200&cid=CFUESB43G) for the ticket, before we consider introducing validation with `libphonenumber`.

PR #824, opened 2019-03-04
LGA-529 - Enable TLS using new cloud platform spec

said-moj:

Enable TLS in the Ingress resource by adding a new certificate resource.

This was implicitly been done by the cloud platforms team before however recent changes mean that we need to explicitly do this.

The circleci service account needed extra permission to be able to create the certificate resource which was cover in this PR https://github.com/ministryofjustice/cloud-platform-environments/pull/501

### DIFF
--- a/kubernetes_deploy/production/ingress.yml
+++ b/kubernetes_deploy/production/ingress.yml
@@ -4,6 +4,9 @@ metadata:
   name: laa-cla-public
   namespace: laa-cla-public-production
 spec:
+  tls:
+  - hosts:
+    - laa-cla-public-production.apps.cloud-platform-live-0.k8s.integration.dsd.io
   rules:
   - host: laa-cla-public-production.apps.cloud-platform-live-0.k8s.integration.dsd.io
     http:
--- a/kubernetes_deploy/staging/ingress.yml
+++ b/kubernetes_deploy/staging/ingress.yml
@@ -4,6 +4,9 @@ metadata:
   name: laa-cla-public
   namespace: laa-cla-public-staging
 spec:
+  tls:
+  - hosts:
+    - laa-cla-public-staging.apps.cloud-platform-live-0.k8s.integration.dsd.io
   rules:
   - host: laa-cla-public-staging.apps.cloud-platform-live-0.k8s.integration.dsd.io
     http:


kubernetes_deploy/staging/ingress.yml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: laa-cla-public
   namespace: laa-cla-public-staging
 spec:
+  tls:
+  - hosts:
+    - laa-cla-public-staging.apps.cloud-platform-live-0.k8s.integration.dsd.io
```
sldblog: 💭 Do we need TLS for staging?

kubernetes_deploy/production/ingress.yml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: laa-cla-public
   namespace: laa-cla-public-production
 spec:
+  tls:
+  - hosts:
+    - laa-cla-public-production.apps.cloud-platform-live-0.k8s.integration.dsd.io
```
sldblog: 🙋‍♂️ We have checklegalaid.service.gov.uk as a public domain as well. Could we add that to the list, please?

said-moj: We don't have checklegalaid.service.gov.uk in the ingress rules, should we still add this under the tls hosts

sldblog: As far as I understand, we need to put the hosts needing TLS certificates here. Good question!
What do you think, @alkar?

kubernetes_deploy/staging/ingress.yml
alkar: Yes! TLS all the things please.

kubernetes_deploy/production/ingress.yml
alkar: I'd say you could leave it like this for now and we can add the service.gov.uk domains when we get to do the certs as well and update the Ingress hosts. This change we asked for is only to ensure a smooth transition onto the new nginx-ingress setup we have.

sldblog: I see, thank you. Do we anticipate any change in the configuration?

alkar: The future changes needed are similar to what we've done in https://github.com/ministryofjustice/fala/pull/32. If you'd like to go ahead and make these changes now, could you drop us a request to update the [config on our side](https://github.com/ministryofjustice/cloud-platform-infrastructure/blob/master/terraform/cloud-platform-components/cert-manager-external.tf#L4-L6)?

said-moj: @sldblog I have added checklegalaid.service.gov.uk to ingress rules and as a tls host

kubernetes_deploy/production/certificate.yml
```diff
@@ -0,0 +1,18 @@
+---
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Certificate
+metadata:
+  name: fala
```
alkar: I think you missed the name here! :)

kubernetes_deploy/production/ingress.yml
```diff
@@ -4,6 +4,13 @@ metadata:
...
+  - hosts:
```
sldblog: To double check, this structure translates to
```json
{
  "spec": {
    "tls": [
      {
        "hosts": [
          "laa-cla-public-production.apps.cloud-platform-live-0.k8s.integration.dsd.io"
        ]
      },
      {
        "hosts": [
          "checklegalaid.service.gov.uk"
        ],
        "secretName": "laa-cla-public-production-tls-certificate"
      }
    ]
  }
}
```
in JSON. Is this the structure we want?

kubernetes_deploy/production/certificate.yml
```diff
@@ -0,0 +1,18 @@
...
+  issuerRef:
+    name: letsencrypt-production
+    kind: ClusterIssuer
+  commonName: 'checklegalaid.service.gov.uk'
+  acme:
+    config:
+    - domains:
+      - 'checklegalaid.service.gov.uk'
+      dns01:
+        provider: route53-dsd
```
sldblog: :neckbeard: No newline at end of file

said-moj:
After thinking about this, I will remove the certificate resource addition to de-risk this and put the addition of the cert resource as a separate PR in the future

So this PR will just be the initial commit https://github.com/ministryofjustice/cla_public/pull/824/commits/1c7899eb39740acc5f71f85886ce19c3931d4976

alkar:
@said-moj that's very reasonable. As discussed the other day, this PR currently includes changes for two different tasks:

- updating the `Ingress` resource with a `tls` field so it will keep working after we migrate to the in-cluster TLS solution
- introducing a `Certificate` and defining the public-facing hostname on the `Ingress` (https://github.com/ministryofjustice/cloud-platform/issues/685)

Splitting them in 2 separate PRs makes sense.

PR #828, opened 2019-03-14
LGA-438 Add pip check to CI

farrepa:

Add `pip check` to CI test job to verify dependency compatibility.

Intentionally left blank.

## Checklist

- [X] Provided JIRA ticket number in the title, e.g. "LGA-152: Sample title"

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,6 +106,10 @@ jobs:
           command: |
             source env/bin/activate
             pip install --requirement requirements.txt --requirement requirements/test.txt
+      - run:
+          name: Check dependencies compatibility
+          command: |
+            source env/bin/activate
             pip check
       - save_cache:
           key: pip-v2-{{ checksum "requirements/base.txt" }}-{{ checksum "requirements/test.txt" }}


sldblog:
Hmm, what happened to #745?

.circleci/config.yml
```diff
@@ -106,6 +106,10 @@ jobs:
           command: |
             source env/bin/activate
             pip install --requirement requirements.txt --requirement requirements/test.txt
+      - run:
+          name: Check dependencies compatibility
```
sldblog: I feel there is sufficient clarity in the `check` output to avoid creating a separate job step. What do you think?

If we vote to keep it, I feel “compatibility” does not reveal that we want to discover broken dependency trees (conflicting versions required with no resolution possible).

farrepa: I don't mind. I slightly prefer the separate job step for clarity. But we should consistently combine or separate across all five repos.

So, consolidate job steps or rename to "Check for conflicting dependencies"?

farrepa:
Discussed with Said, will keep consolidated and amend other PRs.

PR #829, opened 2019-03-14
LGA-487 Edit service down page that shows when cla backend api is down

Katy600:
[LGA-487 Service offline page when CLA Public can't communicate with CLA Backend API](https://dsdmoj.atlassian.net/browse/LGA-541l)

- It edits the service offline pages with new content if the backend is down, to alert users appropriately and guide them on their alternative options.

I was able to check that this service offline page was being displayed by:

- setting DEBUG = False
- running cla_public with out backend
- placing a '/start' at the end of the url generated in the terminal when run locally. 

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/templates/errors/5xx.html
+++ b/cla_public/templates/errors/5xx.html
@@ -2,22 +2,35 @@
 
 {% import "macros/element.html" as Element %}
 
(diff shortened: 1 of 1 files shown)


Katy600:
Screen Shot 2019-03-14 at 10 41 14

PR #830, opened 2019-03-19
LGA-444 Lookup providers from multiple categories

farrepa:

Convert LAALAA_PROVIDER_CATEGORIES_MAP values to lists
Query LAALAA with each category
Concatenate results

Intentionally left blank.

## Checklist

- [X] Provided JIRA ticket number in the title

### DIFF
--- a/cla_public/libs/laalaa.py
+++ b/cla_public/libs/laalaa.py
@@ -58,7 +58,6 @@ def find(postcode, categories=None, page=1):
     merged_data = {"results": []}
     for category in categories:
         data = laalaa_search(postcode=postcode, category=category, page=page)
-        data["results"] = map(decode_categories, data.get("results", []))
-        merged_data["results"].extend(data["results"])
+        merged_data["results"].extend(map(decode_categories, data.get("results", [])))
         merged_data["origin"] = data.get("origin")
     return merged_data
--- a/cla_public/libs/laalaa.py
+++ b/cla_public/libs/laalaa.py
@@ -54,10 +54,11 @@ def decode_categories(result):
     return result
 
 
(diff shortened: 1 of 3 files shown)


cla_public/libs/laalaa.py
```diff
@@ -54,10 +54,11 @@ def decode_categories(result):
...
-
-    data["results"] = map(decode_categories, data.get("results", []))
-
-    return data
+def find(postcode, categories=None, page=1):
+    merged_data = {"results": []}
+    for category in categories:
+        data = laalaa_search(postcode=postcode, category=category, page=page)
+        data["results"] = map(decode_categories, data.get("results", []))
+        merged_data["results"].extend(data["results"])
```
exonian: As we're just extending the results into a new dict's `results` (rather than returning the original dict, modified), there doesn't seem any point doing this in two steps (putting the mapped results into the dict, then getting them from there to extend the merged dict)

cla_public/libs/laalaa.py
```diff
@@ -54,10 +54,11 @@ def decode_categories(result):
...
+        merged_data["origin"] = data.get("origin")
```
exonian: Just checking we don't need the other keys that `laalaa_search` returns? (`next`, `previous`, `count`)

farrepa: Indeed we don't 🚮

cla_public/libs/laalaa.py
farrepa: Less ugly than I thought it would be.

exonian: Ah it was aesthetics. Fair. I figured it was just a hangover from the previous pattern :)

PR #836, opened 2019-03-28
LGA-590 - Adding certificate resource to deploy to work with custom domain

said-moj:
Introducing a Certificate and defining the public-facing hostname on the Ingress
Migrating to the in-cluster TLS solution by updating the Ingress resource with a tls field

This was implicitly being done by the cloud platforms team before however recent changes mean that we need to explicitly do this.

## Checklist


kubernetes_deploy/production/ingress.yml
```diff
@@ -15,3 +19,10 @@ spec:
         backend:
           serviceName: laa-cla-public
           servicePort: 80
+  - host: checklegalaid.service.gov.uk
+    http:
```
sldblog: _not a blocker_
Is there a way to re-use the previous definition? With alternative syntax or YAML anchors?

PR #840, opened 2019-04-16
LGA-599 & LGA-603 Feedback position

joshlow:

-- adds a link to feedback survey where it was missing 
-- replaces links to survey with GDS link style
-- adds a finish button to the confirmation page, instead of the survey banner, so user can exit the service
-- adds GA tracking to the finish button to monitor responses

I couldn't get the GA tracking to work on the staging site, so I'll have to check the configuration is sending events to analytics on the live site.

## Checklist

Relates to Jira tickets #599 and #603

### DIFF
--- a/cla_public/templates/checker/result/confirmation.html
+++ b/cla_public/templates/checker/result/confirmation.html
@@ -178,6 +178,7 @@
   </script>
 
   <script>
+    ga('send', 'event', 'confirmation', 'survey', '{{ _('Finish') }}');
     ga('send', 'event', 'confirmation', 'contact-type', '{{ session.stored.contact_type }}');
     {% if session.stored.adaptations %}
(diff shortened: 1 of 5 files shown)


exonian:
Could you include the JIRA ticket numbers in the PR title, like these ones please? https://github.com/ministryofjustice/cla_public/pulls?q=is%3Apr+is%3Aclosed

exonian:
Are there any uses of `{{ Element.satisfaction_survey_prompt() }}` still left? If there aren't, we can remove the macro, but obviously if there are we need to leave it in.

joshlow:
Good point. There aren't any macros left (that I can see).

exonian:
I'd actually be tempted to make that macro produce the code you're now using in a few places, then, but as discussed yesterday we can do that at a later date if need be

joshlow:
The macro sounds good - shall i make a ticket? Maybe I can pair with you :)

exonian:
Yes to both!

PR #844, opened 2019-05-14
Shrink Docker image

sldblog:

Switches the Docker image to use Alpine Linux and pre-installed nginx and uwsgi. This reduces the image size by about ~350MB:

```
cla_public$ docker images | grep cla-public
.../laa-get-access/laa-cla-public     shrink-image.9265fb3          589MB
.../laa-get-access/laa-cla-public     master                        961MB
```

## Why do these changes?

The reduced image size will

- speed up running the end-to-end tests
- speed up deployments (images will be pulled down faster)

The main benefit (I think) though is the removal of reliance on the previous base image as now the process tree is in a single command and single file (`supervisord-services.ini`).

Since the image replaced the `/sbin/my_init` supervision process, `supervisord` was introduced as a replacement. Consequently, the `*.service` files were replaced with a single `supervisord-services.ini`.

## Checklist

- [ ] Provided JIRA ticket number in the title, e.g. "LGA-152: Sample title" **Not applicable.**
- [x] Related to https://github.com/ministryofjustice/cla_backend/pull/573
- [x] End-to-end tests passed ✅


Dockerfile
```diff
@@ -1,79 +1,50 @@
...
+      zlib-dev
 
-# Set timezone
-RUN ln -fs /usr/share/zoneinfo/Europe/London /etc/localtime
-
-# Install nginx
-RUN add-apt-repository ppa:nginx/stable && \
-    apt-get update && \
-    apt-get -y --force-yes install nginx-full && \
-    chown -R www-data:www-data /var/lib/nginx && \
```
sldblog: :no_entry: Alpine's nginx is using `/var/tmp/nginx` and it should be reflected in the pull request.

sldblog:
I am unable to follow-up on this pull request.
Closing, but leaving the branch for future reference.

sldblog:
@said-moj @exonian

Dockerfile
```diff
@@ -1,79 +1,50 @@
...
-
-ENV HOME /root
-
-# Install python and build packages
-RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && \
-    apt-get -y --force-yes install \
-      apt-utils \
-      build-essential \
-      g++ \
+FROM alpine:3.9
```
sldblog: Can be alpine:3.10 now

PR #851, opened 2019-06-27
LGA-640, LGA-631 - Adding monitoring of activity

said-moj:

Warn users that their session is about to time out and let them extend it if possible.

Incorporating GDS components by coping the following files from https://github.com/hannalaakso/accessible-timeout-warning as instructed in the repo README:
- cla_public/static-src/javascripts/modules/modal-dialog/modal-dialog.js
- cla_public/static-src/javascripts/modules/modal-dialog/dialog-polyfill.0.4.3.js
- cla_public/static-src/stylesheets/_modal-dialog.scss
- cla_public/templates/_modal-dialog.html

For more information about GDS timeout component see https://github.com/alphagov/govuk-design-system-backlog/issues/104

Made a few changes to the GDS component so that we can make our overrides without putting our changes inside the component. I have Submitted these as a PR:
https://github.com/hannalaakso/accessible-timeout-warning/pull/2
https://github.com/hannalaakso/accessible-timeout-warning/pull/3

## Checklist

### DIFF
--- a/cla_public/apps/base/views.py
+++ b/cla_public/apps/base/views.py
@@ -154,6 +154,7 @@ def show_session():
 
 @base.route("/session-expired")
 def session_expired():
+    session.clear()
     return render_template("session-expired.html")
 
 
--- a/cla_public/static-src/javascripts/modules/modal-dialog.js
+++ b/cla_public/static-src/javascripts/modules/modal-dialog.js
@@ -1,8 +1,0 @@
- 'use strict';
-
-  moj.Modules.modalDialog = {
-    init: function() {
-      window.GOVUK.modalDialog.init();
-    }
-  };
-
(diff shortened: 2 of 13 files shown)


cla_public/apps/base/views.py
```diff
@@ -154,6 +154,7 @@ def show_session():
 
 @base.route("/session-expired")
 def session_expired():
+    session.clear()
```
exonian: Was it a mistake that this wasn't there before, or is it newly required because we're redirecting from the front end now?

said-moj: We are redirecting to it from the front-end now. So we added have a 5 minutes countdown and if the user doesn't cancel they're redirected to this page.

I think this page should end the session if hasn't already as it's showing the session-expired page to the user.

exonian: Good point

PR #860, opened 2019-08-22
LGA 752 remove survey

joshlow:

This PR replaces the Survey Monkey feedback link in the header and the Zendesk link in the footer with the Contact CLA feedback route. This gives users a consistent way of getting help if they have a problem using the service.

This PR also includes changes to the feedback form (using zendesk) previously linked from the footer. We are proposing future changes to the feedback route and may use the changed form to capture details if users are experiencing technical issues with the service.

## Checklist
 Related ticket is: https://dsdmoj.atlassian.net/browse/LGA-752

### DIFF
--- a/cla_public/static-templates/errors/50x.html
+++ b/cla_public/static-templates/errors/50x.html
@@ -111,7 +111,7 @@
       <section id="content" role="main">
         <div class="phase-banner">
           <p>
-            <span id="aria-feedback"><a href="https://checklegalaid.service.gov.uk/reasons-for-contacting">Report a problem</a> if you need help or would like to give feedback to improve this service</span>
+            <span id="aria-feedback"><a href="/reasons-for-contacting">Report a problem</a> if you need help or would like to give feedback to improve this service</span>
           </p>
         </div>
         <div class="inner-block ">
--- a/cla_public/static-templates/errors/50x.html
+++ b/cla_public/static-templates/errors/50x.html
@@ -111,7 +111,7 @@
       <section id="content" role="main">
         <div class="phase-banner">
           <p>
-            <span id="aria-feedback">Your <a href="/feedback" aria-labelledby="aria-feedback">feedback</a> will help us continue to improve this service.</span>
+            <span id="aria-feedback"><a href="https://checklegalaid.service.gov.uk/feedback">Contact us</a> if you need help or would like to give feedback to improve this service</span>
           </p>
         </div>
         <div class="inner-block ">
--- a/cla_public/static-templates/errors/50x.html
+++ b/cla_public/static-templates/errors/50x.html
@@ -111,7 +111,7 @@
       <section id="content" role="main">
         <div class="phase-banner">
           <p>
-            <span id="aria-feedback"><a href="https://checklegalaid.service.gov.uk/feedback">Contact us</a> if you need help or would like to give feedback to improve this service</span>
+            <span id="aria-feedback"><a href="https://checklegalaid.service.gov.uk/reasons-for-contacting">Report a problem</a> if you need help or would like to give feedback to improve this service</span>
           </p>
         </div>
         <div class="inner-block ">
@@ -135,7 +135,7 @@
             <ul>
               <li><a href="/privacy">Privacy Policy</a></li>
               <li><a href="/cookies">Cookies</a></li>
-              <li><a href="/feedback">Feedback</a></li>
+              <li><a href="/reasons-for-contacting">Contact us</a></li>
               <li>
                 Built by
                 <a href="https://mojdigital.blog.gov.uk/"><abbr title="Ministry of Justice">MOJ</abbr> Digital Services</a>
--- a/cla_public/templates/_phase-tag.html
+++ b/cla_public/templates/_phase-tag.html
@@ -13,7 +13,7 @@
 
   <p>
     <span id="aria-feedback">
-      <a href="/reasons-for-contacting">Contact us</a> if you need help or would like to give feedback to improve this service
+      <a href="{{ url_for('base.reasons_for_contacting')}}">Contact us</a> if you need help or would like to give feedback to improve this service
   </p>
 
 </div>
--- a/cla_public/templates/_phase-tag.html
+++ b/cla_public/templates/_phase-tag.html
@@ -13,14 +13,7 @@
 
   <p>
     <span id="aria-feedback">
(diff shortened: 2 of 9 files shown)


cla_public/templates/base.html
```diff
@@ -129,5 +129,5 @@
   <li><a href="{{ url_for('base.online_safety') }}">{{ _('Staying safe online') }}</a></li>
   <li><a href="{{ url_for('base.privacy') }}">{{ _('Privacy Policy') }}</a></li>
   <li><a href="{{ url_for('base.cookies') }}">{{ _('Cookies') }}</a></li>
-  <li><a href="{{ url_for('base.feedback') }}">{{ _('Feedback') }}</a></li>
+  <li><a href="https://checklegalaid.service.gov.uk/reasons-for-contacting">{{ _('Contact') }}</a></li>
```
sldblog: :raising_hand_man: It's preferable to use `url_for` than to use a full domain URL as we can configure the former to work on other environments (currently all environments would use production for this link).

cla_public/static-templates/errors/50x.html
```diff
@@ -111,7 +111,7 @@
       <section id="content" role="main">
         <div class="phase-banner">
           <p>
-            <span id="aria-feedback">Your <a href="/feedback" aria-labelledby="aria-feedback">feedback</a> will help us continue to improve this service.</span>
+            <span id="aria-feedback"><a href="https://checklegalaid.service.gov.uk/reasons-for-contacting">Report a problem</a> if you need help or would like to give feedback to improve this service</span>
```
sldblog: :raising_hand_man: We can skip the scheme and the domain if we are linking to a page in this application. This will ensure it works across all environments (development, end-to-end tests, staging, etc.)
```suggestion
            <span id="aria-feedback"><a href="/reasons-for-contacting">Report a problem</a> if you need help or would like to give feedback to improve this service</span>
```

cla_public/templates/base.html
joshlow: Thanks for the review and comments @sldblog 

What would be the correct way of writing this? 

Can you make a suggestion like the last comment. That was really helpful way of committing.

cla_public/templates/_phase-tag.html
```diff
@@ -13,14 +13,7 @@
...
     <span id="aria-feedback">
-      Tell us about your experiences. Your
-      <a
-        rel="noopener noreferrer"
-        target="_blank"
-        href="https://www.surveymonkey.co.uk/r/LegalInfoJourneyMapping_CLA"
-        aria-labelledby="aria-feedback">
-        feedback<span class="visuallyhidden">(Opens in a new window)</span></a>
-        will help us continue to improve this service.</span>
+      <a href="https://checklegalaid.service.gov.uk/reasons-for-contacting">Contact us</a> if you need help or would like to give feedback to improve this service
```
sldblog: Oops, missed this one
```suggestion
      <a href="/reasons-for-contacting">Contact us</a> if you need help or would like to give feedback to improve this service
```

cla_public/templates/base.html
sldblog: It sounds like we want to use `url_for('base.reasons_for_contacting')` to return `/reasons-for-contacting` everywhere as this template already exists:

https://github.com/ministryofjustice/cla_public/blob/bafcec056500df246bb8ba7df71c3bece3f18f64/cla_public/templates/reasons-for-contacting.html

@exonian and @said-moj can tell if I'm talking sense or not :joy:

cla_public/templates/feedback.html
```diff
@@ -3,7 +3,7 @@
 {% import "macros/form.html" as Form %}
 {% import "macros/element.html" as Element %}
 
-{% set title = _('Your feedback') %}
+{% set title = _('Report a problem') %}
```
sldblog: Do we still use this template? It sounds like we're using https://github.com/ministryofjustice/cla_public/blob/bafcec056500df246bb8ba7df71c3bece3f18f64/cla_public/templates/reasons-for-contacting.html instead. (We may want to set up a redirect from /feedback to /reasons-for-contacting?)

cla_public/templates/base.html
joshlow: Makes sense! Will make the change.

joshlow:
@sldblog - I've made the changes, can you review again when you get a spare moment. Thanks!

cla_public/templates/_phase-tag.html
```diff
@@ -13,14 +13,7 @@
...
     <span id="aria-feedback">
-      Tell us about your experiences. Your
-      <a
-        rel="noopener noreferrer"
-        target="_blank"
-        href="https://www.surveymonkey.co.uk/r/LegalInfoJourneyMapping_CLA"
-        aria-labelledby="aria-feedback">
-        feedback<span class="visuallyhidden">(Opens in a new window)</span></a>
-        will help us continue to improve this service.</span>
+      <a href="/reasons-for-contacting">Contact us</a> if you need help or would like to give feedback to improve this service
```
said-moj: Can `{{ url_for('base.reasons_for_contacting')}}` be used here instead of `/reasons-for-contacting`?

joshlow: Hey @said-moj - i've made the change, could you review again when you get the chance?

cla_public/translations/cy/LC_MESSAGES/messages.po
```diff
@@ -48,16 +48,16 @@ msgstr "Na, llenwais y ffurflen hon fy hun"
 
 #: cla_public/apps/base/constants.py:19
 msgid ""
-"I have difficulty using computers so someone filled in this form for me"
-msgstr "Rwyf yn cael trafferth defnyddio cyfrifiaduron, felly llenwodd rhywun arall y ffurflen hon ar fy rhan"
+"I had help filling in this form"
+msgstr ""
```
exonian: Is there a reason the other two have kept their approximate translations but this one now has none? Was it decided that it was too different from "I had some other kind of help"?

exonian:
Making the edits like this looks fine to me 👍

cla_public/translations/cy/LC_MESSAGES/messages.po
joshlow: @exonian - good spot, that's an oversight on my part, will add it back in.

PR #861, opened 2019-08-23
LGA-751 Fixing form submitting on Edge browser

exonian:

Rewrites ajax form validation to be done on a POST request, rather than an OPTIONS request. Therefore the views have to be able to handle the ajax and non-ajax ways of handling page redirects and displaying errors.

In fixing the behaviour in Edge, we've had to completely change the views to use POST for the ajax request (rather than OPTIONS), so this doesn't just affect Edge, where the view were already not working, but all browsers.

Impacts the following paths, with javascript enabled and disabled:
- `/feedback`
- `/reasons-for-contacting`
- `/result/eligible`
- `/result/provisional`
- `/result/confirmation`
- `/contact`
- `/about`
- `/benefits`
- `/additional-benefits`
- `/property`
- `/savings`
- `/income`
- `/outgoings`
- `/review`

Shouldn't need testing in more than one browser on more than one flow, as the different views share the same front-end code but the back-end code varies more.

## Checklist

### DIFF
--- a/cla_public/apps/base/forms.py
+++ b/cla_public/apps/base/forms.py
@@ -8,12 +8,7 @@ from flask.ext.babel import lazy_gettext as _, get_translations
 from wtforms import TextAreaField, RadioField, SelectMultipleField, StringField, widgets
 from wtforms.validators import InputRequired, Length
 
-from cla_public.apps.base.constants import (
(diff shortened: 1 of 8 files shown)


said-moj:
Looks fine but the end-to-end tests seem to be failing when run against this branch

PR #866, opened 2019-09-17
LGA-800 LGA-799 - Adding cookie consent banner and management

said-moj:
Currently the project uses https://github.com/alphagov/govuk_frontend_toolkit which has been deprecated for https://github.com/alphagov/govuk-frontend.

Instead of creating our own cookie banner we have decided to use the cookie-banner component from https://github.com/alphagov/govuk_publishing_components which depends on the new govuk-frontend

- Inlining base template from jinja-moj-template package so that we can change cookie popup message.

- Adding npm dependencies for govuk-frontend  and govuk_publishing_components
- Include cookie-banner and dependent sass files to our main.scss
- Remove moj_template/static/javascripts/govuk-templete.js as that contained the old cookie popup logic and was not extensible. I have moved the other non cookie related logic into moj js modules.
- Added webpack alias for govuk_publishing_components so that components can be easily required in the moj js modules

Intentionally left blank.

## Checklist


cla_public/templates/base.html
```diff
@@ -51,32 +54,46 @@
 
 {% block head %}
   <meta http-equiv="content-language" content="{{ request.cookies.get('locale', 'en')[:2] }}">
+{%- endblock %}
```
exonian: This looks like a typo and I'm confused why it doesn't break something!

Oh, just seen that there was already one of these in the template lower down. Is it a thing? Or silently ignored? Or is our template structure not what we think it is?

cla_public/templates/base.html
```diff
@@ -51,32 +54,46 @@
...
-    ga('create', '{{ config.GA_ID }}', 'auto');
-    ga('set', 'anonymizeIp', true);
+    if (window.GOVUK.checkConsentCookieCategory('', 'usage')) {
+      ga('create', '{{ config.GA_ID }}', 'auto');
+      ga('set', 'anonymizeIp', true);
+    }
+
+    if (window.GOVUK.checkConsentCookieCategory('', 'brexit')) {
+      // GDS CROSS DOMAIN ANALYTICS
+      ga('create', 'UA-145652997-1', 'auto', 'govuk_shared', {'allowLinker': true});
```
exonian: Not ideal for this to be hard-coded but I guess it's not the end of the world at all

cla_public/templates/base.html
said-moj: `{%-` is apparently for controlling vertical space https://stackoverflow.com/questions/11813422/what-does-this-in-jinja2-template-engine-do
https://tedboy.github.io/jinja2/templ6.html

That was already there, I think it looks like an addition due to how the diff is generated

exonian: Ah ok, didn't realise that was a thing. Cool. Thanks

cla_public/static-src/javascripts/modules/cookie-banner.js
```diff
@@ -0,0 +1,7 @@
+ 'use strict';
+  require('govuk_publishing_components/components/cookie-banner.js');
+  require('govuk_publishing_components/lib/cookie-functions.js');
+
+
+
+
```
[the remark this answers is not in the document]
said-moj: I have updated this, not sure why it's not showing as outdated https://github.com/ministryofjustice/cla_public/pull/866/commits/49d1f39381e31f131812094d060e97f2a16a8c91

PR #869, opened 2019-10-04
LGA-809 - Remove 'Is it safe for us to leave a message on this number?' fields from contact forms

said-moj:

- Remove 'Is it safe for us to leave a message on this number?' fields from 'call me back' and  'Call someone else instead of me' forms
- Remove voice messages text from email templates
- Send SAFE value in safe_to_contact field to the backend when user has requested a callback

We have decided to remove this field due to receiving complaints from end user about not getting voice messages and the call centre provider not being allowed to leave voice messages

## Checklist

### DIFF
--- a/cla_public/apps/contact/forms.py
+++ b/cla_public/apps/contact/forms.py
@@ -70,12 +70,6 @@ class CallBackForm(BabelTranslationsFormMixin, NoCsrfForm):
             Length(max=20, message=_(u"Your telephone number must be 20 " u"characters or less")),
         ],
     )
-    safe_to_contact = RadioField(
-        _(u"Is it safe for us to leave a message on this number?"),
-        choices=CONTACT_SAFETY,
-        default="",  # Backend doesn't accept `None` as valid value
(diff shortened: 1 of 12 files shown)


cla_public/apps/contact/forms.py
```diff
@@ -174,14 +162,15 @@ def process_selected_time(form_time):
             local = local_tz.localize(naive)
             return local.astimezone(pytz.utc).isoformat()
 
+        safe_to_contact = CONTACT_SAFETY[0][0] if self.contact_type.data == CONTACT_PREFERENCE[1][0] else ""
```
exonian: This is pretty hard to read. I think I would assign the values from the constants to meaningfully named variables first, to make it much clearer what's being done

said-moj: I think i'll convert them to extended_choices.Choices so that we can do CONTACT_SAFETY.SAFE. There isn't much use of it so updating it wouldn't be a hassle

PR #874, opened 2019-10-24
Persist query parameters during locale redirect

said-moj:

Persist query parameters during locale redirect

## Checklist

### DIFF
--- a/cla_public/apps/base/__init__.py
+++ b/cla_public/apps/base/__init__.py
@@ -1,4 +1,5 @@
 "Base app"
+import urllib
 import datetime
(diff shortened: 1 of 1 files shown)


exonian:
This looks good. It would be thorough to check whether there's already a querystring and append to it (with a `&` rather than a `?`) if there is, but that might be overkill here.

said-moj:
> check whether there's already a querystring and append to it (with a `&` rather than a `?`)

Not sure I follow this
`request.path` always contains the path without the querystring and request_args is a dict of the parameters without the `?`

exonian:
Oh wow, yeah, I was being foolish. Thanks for explaining!

PR #884, opened 2019-11-18
LGA-654 Update privacy for HGS sole operation

joshlow:
Changes to privacy HTML to delete references to Agilisys and reflect HGS sole operation.

Deletes references to Agilisys.

## Checklist

LGA-654
https://dsdmoj.atlassian.net/browse/LGA-654


joshlow:
> Do translations need to be updated too?
> 
> Otherwise, looks good to me!

Thanks @sldblog. Good point. Will add the translations to the missing translations list (need to work out what the translation process is before we can start doing them).

PR #892, opened 2019-12-17
LGA-892 - Use OPERATOR_HOURS from cla_common

said-moj:

Use OPERATOR_HOURS from cla_common

Intentionally left blank.

## Checklist

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -11,7 +11,7 @@ PyYAML==3.11
 Werkzeug==0.9.6
 argparse==1.2.1
 blinker==1.3
-git+git://github.com/ministryofjustice/cla_common.git@0.3.3#egg=cla_common==0.3.3
+git+https://github.com/ministryofjustice/cla_common.git@0.3.6#egg=cla_common==0.3.6
 itsdangerous==0.24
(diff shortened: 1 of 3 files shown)


requirements/base.txt
```diff
@@ -11,7 +11,7 @@ PyYAML==3.11
 Werkzeug==0.9.6
 argparse==1.2.1
 blinker==1.3
-git+git://github.com/ministryofjustice/cla_common.git@0.3.3#egg=cla_common==0.3.3
+git+https://github.com/ministryofjustice/cla_common.git@0.3.6#egg=cla_common==0.3.6
```
exonian: Is the change from ssh to https significant? As long as it works locally and on circle, I'm fine with it, just worth checking :)

exonian: OK it's https already on cla_backend so no objection here

said-moj: Good catch. 

I copied it over from cla_backend https://github.com/ministryofjustice/cla_backend/blob/master/requirements/base.txt#L21 and that seems to be using https. And it works locally and circleci. Although I don't think I understand the impact of having one over the other

PR #894, opened 2019-12-17
Your Benefits page - updated styles

MalcolmVonMoJ:

LGA-884
Updates the page to use the new GDS styles for tickboxes.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/templates/macros/form.html
+++ b/cla_public/templates/macros/form.html
@@ -55,7 +55,7 @@
   {% set control_value = kwargs.control_value if kwargs.control_value else '1' %}
 
   <span id="{{ field.id }}-hint" class="govuk-hint">
-    Select all that apply.
+    {{ trans }}Select all that apply{{ endtrans }}
   </span>
   {{ render_field_errors(field.errors) if field and field.errors}}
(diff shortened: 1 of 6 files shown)


cla_public/templates/macros/form.html
```diff
@@ -317,8 +317,11 @@
   {% if input_type == 'radio' %}
     <div class="govuk-radios">
     {% for option in field %}
+      {% if option.data in separated_options %}
+        <div class="govuk-radios__divider">or</div>
+      {% endif %}
       <div class="govuk-radios__item">
-        <input class="govuk-radios__input" id="{{option.id}}" type="radio" name="{{field.id}}" value="{{option.value}}" >
+        <input class="govuk-radios__input" id="{{option.id}}" type="radio" name="{{field.id}}" value="{{option.data}}" >
```
said-moj: Why the change from option.value to option.data and what's the difference

MalcolmVonMoJ: option.value was an error introduced during my last change - so I fixed it here.  Should always have been option.data

PR #906, opened 2020-01-13
Use 'key' from nodes, rather than 'label', when calculating category

exonian:

Uses the new, untranslated, 'key' key from the API response when determining the category. This prevents the lookup from failing when passed a step name in Welsh.

Improves the local k8s script by adding deletion of deployments, so that this doesn't need to be run separately.
Requires https://github.com/ministryofjustice/cla_backend/pull/610

## Checklist


exonian:
Do not merge until the corresponding `cla_backend` change has been released.

PR #919, opened 2020-02-07
LGA-945 - Remove django dependecy

said-moj:

- Requires https://github.com/ministryofjustice/cla_public/pull/917
- Remove django dependecy

The package jinja-moj-template is no longer being maintained and has a dependency on django >= 1.3 so that it can use the static helper in it's template file.  However a mimic for this has already been added (see cla_public/django_to_jinja.py) for the flask app. So this package can be used without django.

To remove this package dependency on the old version of django I have created a no-deps.txt requirements file, that should install it without also installing django

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,8 +120,10 @@ jobs:
             source env/bin/activate
             pip install --requirement requirements.txt --requirement requirements/test.txt
             pip check
+            # Don't do pip check for packages installed with --no-deps as that will complain about missing dependencies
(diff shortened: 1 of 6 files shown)


said-moj:
@exonian Can you review this again for me please. Before it was based on the branch in https://github.com/ministryofjustice/cla_public/pull/917 but since that has now been merged into master, i have remove the commits from that branch and just rebased this branch with master to avoid conflicts I was having

PR #920, opened 2020-02-10
LGA-695 - Add test coverage to project

said-moj:

- Wraps test running in coverage
- Stores the output on circleci as artifacts
- Sends coverage results to coveralls.io
- Adds a coveralls badge to the readme
- Swapped test runner from nosetests to unittest-xml-reporting to generate xml result files to be used by circleci
- Generates and stores test metadata in xml format for CircleCI to monitor
- Added a tearDown method to cleanup after tests that create a temporary flask app and use app.test_request_context. If this is not done then the flask will raise an AssertionError: Popped wrong request context.

Similar to https://github.com/ministryofjustice/cla_backend/pull/583

## Checklist

### DIFF
--- a/cla_public/apps/base/tests/test_healthchecks.py
+++ b/cla_public/apps/base/tests/test_healthchecks.py
@@ -38,7 +38,11 @@ class DiskSpaceHealthcheckTest(unittest.TestCase):
 class BackendAPIHealthcheckTest(unittest.TestCase):
     def setUp(self):
         self.app = app.create_app("config/testing.py")
-        self.app.test_request_context().push()
+        self.ctx = self.app.test_request_context()
+        self.ctx.push()
+
+    def tearDown(self):
+        self.ctx.pop()
 
     @mock.patch("requests.get")
     def test_backend_check_fails_if_request_fails(self, request_mock):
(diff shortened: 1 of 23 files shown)


cla_public/apps/base/tests/test_healthchecks.py
```diff
@@ -38,7 +38,11 @@ def test_disk_space_check_fails_when_less_than_2_percent_space_is_available(self
 class BackendAPIHealthcheckTest(unittest.TestCase):
     def setUp(self):
         self.app = app.create_app("config/testing.py")
-        self.app.test_request_context().push()
+        self.ctx = self.app.test_request_context()
```
exonian: Non-blocking, but we seem to be doing exactly the same thing in lots of places (we already were, of course).
I think it would be a nice design to make a subclass of `unittest.TestCase` that implements the `setUp` and `tearDown` methods once, and then inherit from that elsewhere.

cla_public/apps/checker/tests/test_api_already_saved_error.py
```diff
@@ -42,6 +42,9 @@ def setUp(self):
         self._ctx.push()
         self.client = self.app.test_client()
 
+    def tearDown(self):
```
exonian: Did you just miss this class?

said-moj: Yep, good spot. Should be fixed now

PR #922, opened 2020-02-13
LGA-845 currency per period

MalcolmVonMoJ:

Introduces a new pattern for asking for income per period, where the user can select "per week", "per year" etc.  

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/static-src/javascripts/modules/form-errors.js
+++ b/cla_public/static-src/javascripts/modules/form-errors.js
@@ -129,8 +129,18 @@
         }
         var name = this.id.replace(/^field-/, '');
 
+        var labelField = $this.find('#field-label-' + name);
+
+        if (labelField.parents(".cla-currency-by-frequency").length) {
+          labelField = labelField.parents(".cla-currency-by-frequency").children("legend");
+        }
+
+        var labelText = labelField.text().replace(/if yes, /i,"").trim();
+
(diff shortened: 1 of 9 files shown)


cla_public/static-src/javascripts/modules/form-errors.js
```diff
@@ -129,8 +129,18 @@
         }
         var name = this.id.replace(/^field-/, '');
 
+        var labelField = $this.find('#field-label-' + name);
+
+        if (labelField.parents(".cla-currency-by-frequency").length) {
+          labelField = labelField.parents(".cla-currency-by-frequency").children("legend");
+        }
+
+        var labelText = labelField.text().replace("If yes, ","").replace("If Yes, ","").trim();
```
said-moj: By doing replace twice, are you trying to match upper and lower case "If Yes"

If so you an use a regex to ignore casing when matching:
`replace(/if yes, /i,"") `

cla_public/static-src/javascripts/modules/form-errors.js
```diff
@@ -175,6 +185,8 @@
 
         var label = $('#field-label-' + fieldName);
 
+        if (label.parents(".cla-currency-by-frequency").length) label = label.parents(".cla-currency-by-frequency").children("legend");
```
said-moj: Can this be split out into multiple lines to make readability easier i.e. 
`if (condition) {
...
}`

PR #933, opened 2020-04-06
LGA-1082 quick escape button

MalcolmVonMoJ:

Adds a quick escape button to the service for certain routes:
- Javascript only
- When displayed, always visible
- Fires of a GA function
- Replaces current tab with google and opens BBC weather in a new tab

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/templates/base.html
+++ b/cla_public/templates/base.html
@@ -124,7 +124,7 @@
       $(".laa-flee-button-container").addClass("laa-fee-button-enabled");
     }
 
-    function flee() {
+    function fleeFromPage() {
       $("body").hide();
       {#
         The below are all other possibilities but not necessarily the best approach. 
@@ -138,13 +138,13 @@
       window.location.replace("http://www.google.co.uk");
     }
     $("#quick-exit").on("click", function() {
-      flee();
       ga('moj-send', 'event', 'quick-exit', 'Mouse-click');
+      fleeFromPage();
     });
     $(document).keyup(function(e) {
       if (e.keyCode == 27) { // escape key
-        flee();
         ga('moj-send', 'event', 'quick-exit', 'Escape-key');
+        fleeFromPage();
       }
     });
 
--- a/cla_public/templates/base.html
+++ b/cla_public/templates/base.html
@@ -53,16 +53,16 @@
       {% endfor %}
     {% endif %}
   {% endif %}
-  <div class="laa-flee-button-container" style="float:right;">
+  <div class="laa-flee-button-container">
     <div class="laa-flee-button-position">
-      <button id="quick-exit" class="govuk-button govuk-button--warning laa-flee-button">
+      <button id="quick-exit" class="govuk-button govuk-button--warning govuk-!-margin-bottom-2 laa-flee-button">
         {{ _('Quick exit') }}
       </button>
-      <p class="exit-info govuk-body-s govuk-!-font-size-14">
+      <p class="laa-quick-exit-info govuk-body-s govuk-!-font-size-14">
         {{ _('Click to hide this page') }}
       </p>
     </div>
-    <button class="govuk-button govuk-button--warning laa-flee-button-sizing" style="visibility: hidden;">
+    <button class="govuk-button govuk-button--warning govuk-!-margin-bottom-2 laa-flee-button-sizing" style="visibility: hidden;">
       {{ _('Quick exit') }}
     </button>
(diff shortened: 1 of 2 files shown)


cla_public/templates/base.html
```diff
@@ -84,6 +118,36 @@
...
+      {#
+        The below are all other possibilities but not necessarily the best approach. 
+
+        var histLength = history.length;
+        window.history.go(-histLength);
+        history.pushState({page: 1}, "Cookies", "/cookies")
+        history.pushState({page: 2}, "Bank Holidays in the UK", "/bank-holidays")
+      #}
+      window.open("http://bbc.co.uk/weather", "_newtab");
+      window.location.replace("http://www.google.co.uk");
```
said-moj, line 138: Would the following be suitable here:
```
window.open("http://bbc.co.uk/weather", "_newtab");
window.close()
```
Opens bbc weather in a new tab and closes the current tab. Stops the browser back button being used to find previous site

cla_public/templates/base.html
```diff
@@ -84,6 +118,36 @@
...
   <script>
+
+    var currentLocation = window.location.href;
+    if (currentLocation.indexOf("n43n3") > 0 || currentLocation.indexOf("n43n14/n97") > 0 || currentLocation.indexOf("n88") > 0 || currentLocation.indexOf("n149") > 0 || currentLocation.indexOf("n86") > 0) {
+      $(".laa-flee-button-container").addClass("laa-fee-button-enabled");
+    }
+
+    function flee() {
+      $("body").hide();
+      {#
```
said-moj, line 129: Do these comments get rendered on the page?

cla_public/templates/base.html
```diff
@@ -84,6 +118,36 @@
 
 {% block javascripts %}
   <script>
+
+    var currentLocation = window.location.href;
+    if (currentLocation.indexOf("n43n3") > 0 || currentLocation.indexOf("n43n14/n97") > 0 || currentLocation.indexOf("n88") > 0 || currentLocation.indexOf("n149") > 0 || currentLocation.indexOf("n86") > 0) {
```
said-moj, line 123: Is this matching paths in the domestic abuse category? If so you should be able to read the current category with something like session.stored.category (Add the current category as class on the main content wrapper)

cla_public/templates/base.html
```diff
@@ -84,6 +118,36 @@
...
+      window.location.replace("http://www.google.co.uk");
+    }
+    $("#quick-exit").on("click", function() {
+      flee();
+      ga('moj-send', 'event', 'quick-exit', 'Mouse-click');
+    });
+    $(document).keyup(function(e) {
+      if (e.keyCode == 27) { // escape key
+        flee();
+        ga('moj-send', 'event', 'quick-exit', 'Escape-key');
```
said-moj, line 147: If flee is changing the browser location does the GA code still get executed?
Is it possible to rename flee to something more descriptive, i.e. quick_exit

cla_public/templates/base.html
MalcolmVonMoJ: No

cla_public/templates/base.html
MalcolmVonMoJ: This doesn't work unless the page has been opened with JS.  It is one of the methods we can perhaps look at long term but it will mean a few changes elsewhere in the system.  

Although `window.close()` works when typed directly into the console, try adding it to a button `<button onclick="window.close();">Close Window</button>` and it won't work.

cla_public/templates/base.html
MalcolmVonMoJ: Hmmm... maybe best to move the location of `ga()` to before the `flee()`...

cla_public/templates/base.html
MalcolmVonMoJ: I have used the `session.checker.diagnosis_previous_choices` check elsewhere on the page for this reason.  Is that enough?
Should I use `session.stored.category` instead of `session.checker.diagnosis_previous_choices`?

cla_public/templates/base.html
said-moj: Can we remove the commented out code then if it's not being used

cla_public/templates/base.html
said-moj: Yep, I had the idea but only tested it on the console. Ignore me

cla_public/templates/base.html
MalcolmVonMoJ: Done

PR #934, opened 2020-04-14
LGA-1095 Code improvement

MalcolmVonMoJ:

Moved the python code to extensions.py

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/apps/base/extensions.py
+++ b/cla_public/apps/base/extensions.py
@@ -19,5 +19,4 @@ def is_quick_exit_enabled(nodes, session=None):
         quick_exit_categories = ["n43n3", "n18", "n19", "n88", "n86", "n97", "n149"]
         for category in quick_exit_categories:
             return category in session.checker["diagnosis_previous_choices"]
-                return category
     return False
--- a/cla_public/apps/base/extensions.py
+++ b/cla_public/apps/base/extensions.py
@@ -11,3 +11,13 @@ def asset(filename, min_ext="min"):
         filename = ".".join(amended)
 
     return url_for("static", filename=filename)
+
+
+@base.app_template_global()
+def is_quick_exit_enabled(nodes, session=None):
+    if "diagnosis_previous_choices" in session.checker:
+        quick_exit_categories = ["n43n3", "n18", "n19", "n88", "n86", "n97", "n149"]
+        for category in quick_exit_categories:
+            if category in session.checker["diagnosis_previous_choices"]:
+                return category
+    return False
--- a/cla_public/apps/base/extensions.py
+++ b/cla_public/apps/base/extensions.py
@@ -14,7 +14,7 @@ def asset(filename, min_ext="min"):
 
 
 @base.app_template_global()
-def is_quick_exit_enabled(nodes, session=None):
+def is_quick_exit_enabled(session):
     if "diagnosis_previous_choices" in session.checker:
         quick_exit_categories = ["n43n3", "n18", "n19", "n88", "n86", "n97", "n149"]
         for category in quick_exit_categories:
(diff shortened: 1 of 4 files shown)


cla_public/templates/base.html
```diff
@@ -54,15 +54,9 @@
...
-        ('n97' in session.checker.diagnosis_previous_choices) or
-        ('n86' in session.checker.diagnosis_previous_choices) or
-        ('n149' in session.checker.diagnosis_previous_choices) or
-        ('n18' in session.checker.diagnosis_previous_choices) or
-        ('n19' in session.checker.diagnosis_previous_choices) or
-        ('n88' in session.checker.diagnosis_previous_choices)
-    %}
+    {% if is_quick_exit_enabled(node, session) %}
       laa-fee-button-enabled
+      {{ is_quick_exit_enabled(node, session) }}
```
exonian, line 59: This value has been added to the output in the refactor (ie you've replaced the `if` condition but also added the matching category). Is that intentional? Was it a debugging thing? Or does it achieve some undocumented aim?

cla_public/templates/base.html
```diff
@@ -54,15 +54,9 @@
...
-    {% if ('n43n3' in session.checker.diagnosis_previous_choices) or
-        ('n97' in session.checker.diagnosis_previous_choices) or
-        ('n86' in session.checker.diagnosis_previous_choices) or
-        ('n149' in session.checker.diagnosis_previous_choices) or
-        ('n18' in session.checker.diagnosis_previous_choices) or
-        ('n19' in session.checker.diagnosis_previous_choices) or
-        ('n88' in session.checker.diagnosis_previous_choices)
-    %}
+    {% if is_quick_exit_enabled(node, session) %}
       laa-fee-button-enabled
```
exonian, line 58: I know this isn't new in this PR but this looks like a typo of 'flee'. If that's the case, do you think it's in scope to fix in this PR? If it isn't, it might need a comment.

cla_public/apps/base/extensions.py
```diff
@@ -11,3 +11,13 @@ def asset(filename, min_ext="min"):
...
     return url_for("static", filename=filename)
+
+
+@base.app_template_global()
+def is_quick_exit_enabled(nodes, session=None):
+    if "diagnosis_previous_choices" in session.checker:
+        quick_exit_categories = ["n43n3", "n18", "n19", "n88", "n86", "n97", "n149"]
+        for category in quick_exit_categories:
+            if category in session.checker["diagnosis_previous_choices"]:
+                return category
```
exonian, line 22: If we're returning the category, we might consider renaming the function as the current name suggests a boolean return to me. "Is it enabled?" "n43n3" is a bit unexpected to me. But see the comment in the other file first, for whether we actually need to return the category (the previous version, in the template, didn't)

cla_public/templates/base.html
MalcolmVonMoJ: I shall check and change - I can do it as part of this PR.

cla_public/templates/base.html
MalcolmVonMoJ: A debugging thing...  one of the things I forgot about over Easter...

cla_public/apps/base/extensions.py
```diff
@@ -11,3 +11,13 @@ def asset(filename, min_ext="min"):
...
 
     return url_for("static", filename=filename)
+
+
+@base.app_template_global()
+def is_quick_exit_enabled(nodes, session=None):
+    if "diagnosis_previous_choices" in session.checker:
+        quick_exit_categories = ["n43n3", "n18", "n19", "n88", "n86", "n97", "n149"]
+        for category in quick_exit_categories:
+            if category in session.checker["diagnosis_previous_choices"]:
```
exonian, line 21:
```suggestion
            return category in session.checker["diagnosis_previous_choices"]
```

cla_public/apps/base/extensions.py
exonian, line 22:
```suggestion
```

PR #942, opened 2020-05-11
LGA-956 Error Messages

MalcolmVonMoJ:

- Added specific error messages for each question, removing the format where the question text was repeated with a generic error.  
- Rephrased other error messages.
- New error messages for currency rate question types - errors for no value and no frequency, distinct from each other.  (New variables for this: `freq_message` when amount specified but rate (e.g. per week) not answered, and `amount_message` for the opposite.  This has been added to the MoneyIntervalAmountRequired function in validators.py.)
- Stopped error styling from cascading - error message being present still cascades as before.  
- Error summary re-positioned at the top of the page - JS change and structure change on many pages - this includes the `{{ Form.handle_errors(form) }}` being re-positioned on about 10 pages.
- Error summary changed to not use the dl dt dd structure - as per design system.
- Added new JS function (in base.html) for error summary links so they focus on the errorred answer box but only scrolls as far as the relevant question text.
- Changed the error state code so child questions are not in error state if their parent questions are null - affected pop-up questions when their parent was unanswered, errorred, and subsequently answered.  (This might have been raised as another bug, but I can't find it now.)
- Added Welsh language for new wording.
- Some smart quote correction and white space tweaks.
- New highlighting for error-state answer boxes.
- Errors re-positioned in relation to the question and hint text (in form.html).

- Some smart quote correction.

## Checklist

### DIFF
--- a/cla_public/apps/checker/forms.py
+++ b/cla_public/apps/checker/forms.py
@@ -93,8 +93,8 @@ class AboutYouForm(BaseForm):
         label=_(u"If Yes, how many?"),
         validators=[
             IgnoreIf("have_children", FieldValue(NO)),
-            DataRequired(_(u"Number must be between 1 and 50")),
-            NumberRange(min=1, max=50, message=_(u"Number must be between 1 and 50")),
+            DataRequired(_(u"Tell us how many children you have")),
+            NumberRange(min=1, max=50, message=_(u"Enter a number between 1 and 50")),
         ],
     )
     have_dependants = YesNoField(
--- a/cla_public/apps/checker/forms.py
+++ b/cla_public/apps/checker/forms.py
@@ -191,13 +191,20 @@ class YourBenefitsForm(BaseForm):
         label=_(u"Which benefits do you receive?"),
         partner_label=_(u"Which benefits do you and your partner receive?"),
         choices=BENEFITS_CHOICES,
-        validators=[AtLeastOne()],
+        validators=[AtLeastOne(message=_(u"Select the benefits that you receive"))],
     )
 
     child_benefit = MoneyIntervalField(
         label=_(u"If yes, enter the total amount you get for all your children"),
         choices=money_intervals("", "per_week", "per_4week"),
-        validators=[IgnoreIf("benefits", FieldValueNotIn("child_benefit")), MoneyIntervalAmountRequired()],
+        validators=[
+            IgnoreIf(
+                "benefits",
+                FieldValueNotIn("child_benefit"),
+                message=_(u"Enter the total amount you get for all your children"),
+            ),
+            MoneyIntervalAmountRequired(message=_(u"Tell us how often do you get child benefit")),
+        ],
     )
 
     @classmethod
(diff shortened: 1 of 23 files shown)


cla_public/apps/checker/validators.py
```diff
@@ -73,21 +73,41 @@ def __call__(self, form, field):
...
     def __call__(self, form, field):
         amount = field.form.per_interval_value
+        interval = field.form.interval_period
+        amount_not_set = amount.data is None
+        nonzero_amount = amount.data > 0
+        interval_selected = interval.data != ""
 
-        if not amount.errors and amount.data is None:
-            message = self.message or field.gettext(u"Please provide an amount")
+        if not amount.errors and interval_selected and amount_not_set and self.amount_message:
```
exonian, line 93: This whole method is extremely hard to read now. In particular, I would be interested in making this `if` statement not have four conditions (either by combining them into a variable first or -- probably preferably -- moving one of more other `if` statements before it so that we're never dealing with so many factors all at once (eg if we can do a particular thing in all cases where one particular thing is true, then we might be able to do that early on and then not have to include it in any other cases).

We could also maybe split that intervals stuff out into a method, which would at least make that pathway less intimidating to take in.

Happy for you to work on this yourself if you'd like to, more than happy to pair on it.

cla_public/apps/checker/forms.py
```diff
@@ -456,34 +521,50 @@ def title(self):
...
+                _(u"Tell us how much you pay towards Criminal Legal Aid, or enter 0 if this doesn’t apply to you")
+            )
+        ],
     )
     childcare = PartnerMoneyIntervalField(
         label=_(u"Childcare"),
-        description=_(
-            u"Money you pay for your child to be looked after while " u"you work or study outside of your home"
-        ),
+        description=_(u"Money you pay for your child to be looked after while you work or study outside of your home"),
```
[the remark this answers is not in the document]
MalcolmVonMoJ: There is a difference and it is extremely obscure - the error message text has not been okayed by Josh, so this will be picked up in LGA-940.  Until that is done we will use the default text for this.

cla_public/apps/checker/validators.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, message=None, freq_message=None, amount_message=None):
         self.freq_message = freq_message
         self.amount_message = amount_message
 
-    def append_interval_text(self, field):
+    def append_interval_text(self, field, interval):
```
exonian, line 86: Oh yes, I spotted this when we were writing the signature but planned to lead you towards finding it yourself and then completely forgot 🤦 😅

cla_public/apps/checker/validators.py
```diff
@@ -73,21 +73,43 @@ def __call__(self, form, field):
...
 
 class MoneyIntervalAmountRequired(object):
-    def __init__(self, message=None):
+    def __init__(self, message=None, freq_message=None, amount_message=None):
         self.message = message
+        self.freq_message = freq_message
+        self.amount_message = amount_message
+
+    def append_interval_text(self, field, interval):
+        if interval.data == "per_week":
```
exonian, line 87: On reflection, I think this might be worth refactoring to avoid the multiple `if`s. Do you fancy pairing on it? Think it could be quite enjoyable.

PR #947, opened 2020-05-14
LGA-1158 Escape key activates quick exit only when quick exit visible

MalcolmVonMoJ:

Added condition so quick exit only fires when quick exit enabled (checks for class)

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/templates/base.html
+++ b/cla_public/templates/base.html
@@ -126,10 +126,8 @@
       fleeFromPage("mouse click");
     });
     $(document).keyup(function(e) {
-      if ($('.laa-flee-button-enabled').length) {
-        if (e.keyCode == 27) { // escape key
-          fleeFromPage("escape key");
-        }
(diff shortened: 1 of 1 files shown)


cla_public/templates/base.html
```diff
@@ -126,8 +126,10 @@
       fleeFromPage("mouse click");
     });
     $(document).keyup(function(e) {
-      if (e.keyCode == 27) { // escape key
-        fleeFromPage("escape key");
+      if ($('.laa-flee-button-enabled').length) {
```
exonian, line 129: Given that we're only using one of the logic paths (ie there's no `else` behaviour), would this read more simply if combined into one `if <code> and <enabled>` statement rather than being nested?

MalcolmVonMoJ: Yes, sure - I was thinking at the time that this way was a little more readable, but I guess it isn't really that different.

MalcolmVonMoJ:
I don't think we have any tests in this area.

PR #948, opened 2020-05-21
LGA-1181 COVID-19 opening times changed by JS function

MalcolmVonMoJ:

New JS to hide unwanted days and times.  Done by matching the text in the options.  
Also function to disable "call today" on a weekend.  
To be reversed once LGA-1053 is complete

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/templates/base.html
+++ b/cla_public/templates/base.html
@@ -121,8 +121,8 @@
       $("#thirdparty-time-time_in_day, #thirdparty-time-time_today, #callback-time-time_in_day, #callback-time-time_today")
         .find("option:contains(7:30 PM), option:contains(6:30 PM), option:contains(5:30 PM)")
         .remove();
-      var d = new Date();
-      var n = d.getDay();
+      const d = new Date();
+      const n = d.getDay();
       if (n == 6 || n === 0) {
         /*if we are currently in a weekend*/
         $("#thirdparty-time-specific_day-0, #callback-time-specific_day-0").attr('disabled', 'disabled');
--- a/cla_public/templates/base.html
+++ b/cla_public/templates/base.html
@@ -123,7 +123,7 @@
         .remove();
       var d = new Date();
       var n = d.getDay();
-      if (n > 5) {
+      if (n == 6 || n === 0) {
         /*if we are currently in a weekend*/
         $("#thirdparty-time-specific_day-0, #callback-time-specific_day-0").attr('disabled', 'disabled');
         $("#thirdparty-time-specific_day-1, #callback-time-specific_day-1").prop('checked', true);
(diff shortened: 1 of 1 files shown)


cla_public/templates/base.html
```diff
@@ -113,6 +113,23 @@
 {% block javascripts %}
   <script>
 
+    $(document).ready(function(){
+      /*COVID-19: This function is a stopgap for emergency plague-time opening times*/
```
exonian, line 117:
```suggestion
      /*COVID-19: This function is a stopgap for emergency opening times*/
```

cla_public/templates/base.html
```diff
@@ -113,6 +113,23 @@
...
+        .find("option:contains(Saturday), option:contains(Sunday)")
+        .remove();
+      $("#thirdparty-time-time_in_day, #thirdparty-time-time_today, #callback-time-time_in_day, #callback-time-time_today")
+        .find("option:contains(7:30 PM), option:contains(6:30 PM), option:contains(5:30 PM)")
+        .remove();
+      var d = new Date();
+      var n = d.getDay();
+      if (n > 5) {
+        /*if we are currently in a weekend*/
+        $("#thirdparty-time-specific_day-0, #callback-time-specific_day-0").attr('disabled', 'disabled');
```
mayowaadeniyi, line 128: Is it possible to just hide the "call back on weekend" option completely, so users don't need to wonder why it's disabled?

MalcolmVonMoJ: Yes it is - I originally used `.hide()` instead, but then that leads to the question about whether we need the radio button at all as there is only one option and then that leads to awkwardness in displaying the pop-up questions.  

In the end, and in the absence of Rob, I decided to leave it as unchanged as possible.  It negates any possible layout issues and users can then see that the option is available on other days.

cla_public/templates/base.html
```diff
@@ -113,6 +113,23 @@
...
+      /*COVID-19: This function is a stopgap for emergency opening times*/
+      $("#field-thirdparty-time-day, #field-callback-time-day")
+        .find("option:contains(Saturday), option:contains(Sunday)")
+        .remove();
+      $("#thirdparty-time-time_in_day, #thirdparty-time-time_today, #callback-time-time_in_day, #callback-time-time_today")
+        .find("option:contains(7:30 PM), option:contains(6:30 PM), option:contains(5:30 PM)")
+        .remove();
+      var d = new Date();
+      var n = d.getDay();
+      if (n > 5) {
```
mayowaadeniyi, line 126: I also wanted to quickly point out here Malcolm.

Sunday is given the number 0, so this would work for Saturday, but Sunday would not be counted. Wouldn't this need to be changed to 

> n > 5 && n === 0

>

MalcolmVonMoJ: Changed - using `||` instead of `&&`

PR #950, opened 2020-05-22
LGA-1119 Change GA language field on Welsh site from en-gb to en-cy

mayowaadeniyi:

After switching to the welsh language, when navigating welsh pages, each page will send 'cy_GB' as the language code.

Adds Google Analytics credentials to Kubernetes to enable testing on production and staging. This was outside the scope of the ticket, but it was necessary in order to test my changes which is why it was included.

Sets the fallback English language code as 'en_GB' instead of en'.

## Checklist

### DIFF
--- a/cla_public/templates/base.html
+++ b/cla_public/templates/base.html
@@ -183,7 +183,7 @@
         ga(values["name"] + '.linker:autoLink', ['www.gov.uk']);
       }
       if(localeLanguage === 'cy') {
-        ga(values["name"] + '.set', 'language', 'cy')
+        ga(values["name"] + '.set', 'language', 'en_cy')
       }
     });
 
--- a/cla_public/templates/base.html
+++ b/cla_public/templates/base.html
@@ -171,6 +171,8 @@
       }
     {% endif %}
 
+    const localeLanguage = "{{ request.cookies.get('locale', 'en')[:2] }}"
+
     _.forEach(window.ga_trackers, function(values, id){
       ga('create', values["id"], values["domain"], values["name"], values["options"]);
       ga(values["name"] + '.set', 'anonymizeIp', true);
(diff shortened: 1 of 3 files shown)


cla_public/templates/base.html
```diff
@@ -180,6 +182,9 @@
         ga(values["name"] + '.linker.set', 'anonymizeIp', true);
         ga(values["name"] + '.linker:autoLink', ['www.gov.uk']);
       }
+      if(localeLanguage === 'cy') {
+        ga(values["name"] + '.set', 'language', 'en_cy')
+      }
```
MalcolmVonMoJ: Is `en_cy` correct?  I'd have thought either `cy_GB` or simply `cy`.
Also, this is setting it to Welsh if it is already Welsh?  Does that achieve anything?  I would have thought we'd need to account for the Korean situation, by setting that to English.  
Maybe adding
` else { ga(values["name"] + '.set', 'language', 'en') }`

mayowaadeniyi: Not sure tbh...I used en_cy to match what was asked in requirements. I would be more than happy to change that to cy, I think that would make it easier to understand.

To answer your 2nd question - I had to do it like this because Google Analytics doesn't care about the locale set, it only looks at the language in the browser. So I have to tell manually tell Google Analytics to set the language field as Welsh depending on the locale.

And yes, I have added that in my last commit. This will fix the Korean situation.

exonian:
@mayowaadeniyi Be careful with your PR descriptions! I read this one and then, thinking that you hadn't in fact included the Kubernetes env vars changes in it, suggested to Alex you be involved in it next sprint! So just make sure that, if you do end up doing work that's outside the scope of what one might reasonably expect to be in the PR, you mention it in the description 👍

PR #955, opened 2020-06-15
LGA-1231 - Use cla_backend kubernetes staging deployment

said-moj:

Use cla_backend kubernetes staging deployment

backend_base_uri previosuly had the the trailing slash, the paths are passed with the leading slash this will results in the url becoming scheme://domain.tld//uri  which django will see it as a non-existing path and return a 404.

This is not happening on the current deployment system because nginx is sitting in front of uswgi and reconstructing the url.

## Checklist

### DIFF
--- a/helm_deploy/cla-public/values-staging.yaml
+++ b/helm_deploy/cla-public/values-staging.yaml
@@ -6,7 +6,7 @@ image:
   pullPolicy: IfNotPresent
 
 environment: staging
-backend_base_uri: https://staging-backend.cla.dsd.io/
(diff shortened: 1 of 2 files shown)


helm_deploy/cla-public/values-staging.yaml
```diff
@@ -6,7 +6,7 @@ image:
   pullPolicy: IfNotPresent
 
 environment: staging
-backend_base_uri: https://staging-backend.cla.dsd.io/
```
said-moj, line 9: I think this should not have had the the trailing slash as the paths are passed with the leading slash which results in the the following schema://domain.tld//uri which django will see it as a non-existing path and return a 404.

This wasn't happening on the current deployment system because nginx was sitting in front of uswgi and reconstructing the url.

PR #958, opened 2020-06-18
LGA-1227 Welsh translations copied into public

MalcolmVonMoJ:

For all page titles in the scope selection part of the page, the translations are in CLA Backend.  This copies them to the CLA Public translation list as well.  

A couple of missing translations added.

## Checklist


said-moj:
If these strings have translation in cla_backend we should be able to get the translated strings from cla_backend(or atleast make it give us). Was this considered?

MalcolmVonMoJ:
@said-moj 
No, it wasn't considered - and that might be a good idea.  The only issue might be that backend doesn't use smart quotes - which is used for all translations in CLA Public.  It was that discrepancy that alerted me to the bug in the first place.  
But that notwithstanding, how would we go about referencing the backend text in public?

said-moj:
After looking at the backend logic for this, this is definitely caused by a bug in the cla_backend

1. On /start a new diagnosis object is created and the node titles are translated into the current language and stored on that object

2. The choices are fetched on every request(refresh or navigation) in the current language which translate as expected

3. However the titles are part of the object that was created in step 1.
On every request, this object is loaded from the database and the tile value is extracted in it's original language

MalcolmVonMoJ:
These changes will be easy to revert once we have done LGA-1251

PR #960, opened 2020-06-25
LGA-1246 - Add GTM code staging

said-moj:
- Add GTM code to staging

Intentionally left blank.

## Checklist

- [X] Provided JIRA ticket number in the title, e.g. "LGA-152: Sample title"

### DIFF
--- a/cla_public/templates/base.html
+++ b/cla_public/templates/base.html
@@ -148,7 +148,7 @@
     (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
     new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
     j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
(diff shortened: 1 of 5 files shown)


cla_public/templates/base.html
```diff
@@ -141,19 +141,26 @@
...
 <script>
-{% if config.MOJ_GTM_ID %}
-  if (window.GOVUK.checkConsentCookieCategory('', 'usage')) {
+  function add_gtm() {
     (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
     new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
     j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-    'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-    })(window,document,'script','dataLayer','{{ config.MOJ_GTM_ID }}');
+    'https://www.googletagmanager.com/gtm.js?id='+i+dl+ '&gtm_auth=4KOuCU7aXv3Owhcs7U0ORw&gtm_preview=env-3&gtm_cookies_win=x';f.parentNode.insertBefore(j,f);
```
exonian, line 151: Where is the `gtm_auth` value from? If it's account specific, should it be from config like the account ID?

said-moj: yes it's account specific(although both prod and staging share the same gtm account). i'll move it to a config

PR #976, opened 2020-07-29
LGA-1315 - Fix checks for CLA_ENV on production

exonian:
Fixes multiple checks that the `CLA_ENV` setting is the production setting

The value of the env var on kubernetes is "production" not "prod". This
was the case even before the recent changes to the kubernetes config

The homepage redirection was recently changed to check this setting, and
was incorrectly set based on the other uses, which appear to have been
wrong for a while.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/apps/base/forms.py
+++ b/cla_public/apps/base/forms.py
@@ -57,7 +57,7 @@ class FeedbackForm(Honeypot, BabelTranslationsFormMixin, Form):
 
         environment = current_app.config["CLA_ENV"]
         subject = "CLA Public Feedback"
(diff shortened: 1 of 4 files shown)


MalcolmVonMoJ:
Does this line also need to change in `deployment.py`
`SESSION_COOKIE_SECURE = os.environ.get("CLA_ENV", "") in ["prod", "staging"]`

exonian:
@MalcolmVonMoJ Yep great spot, thanks! I did see that one when I was looking for usage on here and then missed it when I switched to looking locally 🤦

PR #980, opened 2020-08-06
LGA-1319 CSS fix for Firefox bug

MalcolmVonMoJ:

The bug is a firefox-specific layout issue where the currency prefix is mis-aligned (vertically) with the input field.  The absolute positioning in Firefox is done after the padding, rather than ignoring padding as in other browsers.  

Added some Firefox-specific CSS to:
- position the currency correctly using a top value.
- add a white outline to the input box to catch any slight overflow from the positioning.

Intentionally left blank.

## Checklist


said-moj:
@MalcolmVonMoJ what's actual bug in Firefox that this fixes?

PR #984, opened 2020-08-12
LGA-1347 Maintenance page

said-moj:

- Adds maintenance mode environment variable
- Adds a before_request action that when maintenance mode is enabled redirects **ALL traffic** to the /maintenance page.

When not in maintenance mode if someone goes to /maintenance then they will be redirected to /start

## Checklist

### DIFF
--- a/cla_public/apps/base/__init__.py
+++ b/cla_public/apps/base/__init__.py
@@ -1,7 +1,7 @@
 "Base app"
 import urllib
 import datetime
-from flask import Blueprint, after_this_request, request, redirect
+from flask import Blueprint, after_this_request, request, redirect, current_app
 
 base = Blueprint("base", __name__)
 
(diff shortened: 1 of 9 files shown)


exonian:
I'm wondering if there's any merit in trying to redirect back from `/maintenance` when we're not in maintenance mode. I know it's a bit of a pet peeve of mine to be shunted to one of these pages and know that refreshing won't help any, but it might not be a big enough user issue to actually do.

(I'll look at the code properly after standup)

said-moj:
> I'm wondering if there's any merit in trying to redirect back from `/maintenance` when we're not in maintenance mode. I know it's a bit of a pet peeve of mine to be shunted to one of these pages and know that refreshing won't help any, but it might not be a big enough user issue to actually do.
> 
> (I'll look at the code properly after standup)

Wow this is actually a something I haven't considered. Definitely sounds like a bug to me, i'll add.

Good catch

cla_public/apps/base/__init__.py
```diff
@@ -22,3 +22,12 @@ def remember_locale(response):
...
             return response
+
+
+@base.before_app_request
+def detect_maintenance():
+    maintenance_mode = current_app.config.get("MAINTENANCE_MODE", False)
+    if maintenance_mode and request.path != u"/maintenance":
+        return redirect("/maintenance")
+    if not maintenance_mode and request.path == u"/maintenance":
+        return redirect("/start")
```
exonian, line 33: I wonder if this should be `/` so that prod goes to the external service start page

PR #986, opened 2020-08-13
LGA-1321 removed duplicate ID

MalcolmVonMoJ:

Corrected input ID

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/templates/checker/result/_find-legal-adviser.html
+++ b/cla_public/templates/checker/result/_find-legal-adviser.html
@@ -10,7 +10,7 @@
     {{ Form.postcode_input(form.postcode) }}
     {% if category %}
       <div class="govuk-form-group">
-        <input class="govuk-input govuk-input--width-10" id="postcode" type="hidden" name="category" value="{{ category }}" />
+        <input class="govuk-input govuk-input--width-10" id="category" type="hidden" name="category" value="{{ category }}" />
       </div>
     {% endif %}
     <button class="govuk-button" type="submit">


cla_public/templates/checker/result/_find-legal-adviser.html
```diff
@@ -10,7 +10,7 @@ <h2 class="govuk-heading-m">{{ _('Find a legal adviser') }}</h2>
     {{ Form.postcode_input(form.postcode) }}
     {% if category %}
       <div class="govuk-form-group">
-        <input class="govuk-input govuk-input--width-10" id="postcode" type="hidden" name="category" value="{{ category }}" />
+        <input class="govuk-input govuk-input--width-10" id="category" type="hidden" name="category" value="{{ category }}" />
```
mayowaadeniyi, line 13: This code looks different to the one pictured in the ticket on JIRA... Where is the duplication?

MalcolmVonMoJ: The code in Jira was how it was rendered.  The duplication was spread over a couple of files which were then compiled together.

mayowaadeniyi: Ah okay cool. Last note I want to say is do we need to add e.g.`<label for="category">Category</label>` so it can be read by screen readers or is that another issue on it's own? This probably has nothing to do with the ticket (LGTM), but just wanted to ask before approving.

MalcolmVonMoJ: Ordinarily you would be right.  But this is "type=hidden" so isn't visible to the user - AT or otherwise.  It is a bit of hidden code that adds another parameter to the search that the user has selected in a different way (before this screen).   Happy to explain further if you want.

mayowaadeniyi: What does AT mean?

MalcolmVonMoJ: Sorry - Assistive Tech - like screen readers and other things that use well-structured semantic HTML.

PR #989, opened 2020-08-17
LGA-1170 no options bug

MalcolmVonMoJ:

Added JS page reload for no-option instances.
Changed wording of no-options page to say "Loading..."
Added comment about reloading page for non-JS situations.
Changed GA naming so reports can be distinguished.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/templates/base.html
+++ b/cla_public/templates/base.html
@@ -105,7 +105,7 @@
     {% if zeroOptions %}
       // this is set in diagnosis.html
       $( document ).ready(function(){
-        ga('moj.send', 'event', 'Zero-options', 'Zero options instance', 'No options instance: '+stripPII(window.location.href));
+        ga('moj.send', 'event', 'Zero-options', 'Zero options (page reload)', 'No options instance: '+stripPII(window.location.href));
         location.reload();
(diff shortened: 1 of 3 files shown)


cla_public/templates/base.html
```diff
@@ -105,7 +105,8 @@
     {% if zeroOptions %}
       // this is set in diagnosis.html
       $( document ).ready(function(){
-        ga('moj.send', 'event', 'Zero-options', 'Zero options instance', 'No options instance: '+stripPII(window.location.href));
+        ga('moj.send', 'event', 'Zero-options', 'Zero options (page reload)', 'No options instance: '+stripPII(window.location.href));
+        location.reload();
```
said-moj, line 109: This could maybe lead them to getting no options again and be stuck on this page.
Can we redirect them to the start page instead?

MalcolmVonMoJ: Every instance of this is solved by refreshing the page.  I've never had the same error occur twice in succession.  However, if that were to happen, it would just reload and then be fine.  

Going to the start page would end their session, so that's not good...

PR #990, opened 2020-08-19
LGA-1326 Aria Expanded attribute

MalcolmVonMoJ:

Moved the aria-expanded code so it only fires when expanded rather than on every click.

CSS changes for narrowish screens.

## Checklist


mayowaadeniyi (approved):
LGTM 👍

PR #992, opened 2020-08-24
LGA-1333 colours etc on map page

MalcolmVonMoJ:

Distance colour changed to black
Focus colour updated list of solicitors
Focus colour updated for the map controls and links
Map and Satellite buttons updated to mimic govuk buttons (not exactly the same colours to maintain suitable contrast with map)
Cat of law tags changed to a list at the bottom of the details bit
New JS to expand the size of the map to be more suitable to the list (avoiding double-scrolling).
Unused pagination controls bit hidden (dormant, can be resurrected if/when needed)

Max-width increased from 660 px to 960 px

## Checklist

### DIFF
--- a/cla_public/static-src/javascripts/modules/find-legal-adviser.js
+++ b/cla_public/static-src/javascripts/modules/find-legal-adviser.js
@@ -307,3 +307,18 @@
     }
   };
 
+  function setMapHeight() {
+    if ($(window).width() > 460 && $("ul.org-list").height() > 460) {
+      var setHeight = $("ul.org-list").height();
+      if (setHeight > $(window).height()) setHeight = $(window).height();
+      $(".search-results-container").height(setHeight);
+      $(".search-results-list").height(setHeight);
+    }
+  }
+  $("document").ready(function(){
+    setMapHeight();
+  });
+
+  $(".search-results-container").click(function(){
+    setMapHeight();
+  });
--- a/cla_public/static-src/javascripts/modules/find-legal-adviser.js
+++ b/cla_public/static-src/javascripts/modules/find-legal-adviser.js
@@ -308,19 +308,18 @@
   };
 
   function setMapHeight(setHeight) {
-    console.log(setHeight);
     if ($(window).width() > 640) {
-      let listHeight = $("ul.org-list").height();
+      var listHeight = $("ul.org-list").height();
       if (listHeight > setHeight) setHeight = listHeight;
       $(".search-results-container, .search-results-list").height(setHeight);
     }
   }
 
   $("document").ready(function(){
-    $(".search-results-container").data("originalheight", $(".search-results-container").height());
-    setMapHeight(460);
+    var originalHeight = $(".search-results-container").height();
+    setMapHeight(originalHeight);
     $(".search-results-container").click(function(){
-      setMapHeight(460);
+      setMapHeight(originalHeight);
     });
   });
 
--- a/cla_public/static-src/javascripts/modules/find-legal-adviser.js
+++ b/cla_public/static-src/javascripts/modules/find-legal-adviser.js
@@ -1,3 +1,4 @@
+const wideScreen = 641;
  'use strict';
   var _ = require('lodash');
   moj.Modules.FindLegalAdviser = {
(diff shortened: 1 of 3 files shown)


cla_public/static-src/javascripts/modules/find-legal-adviser.js
```diff
@@ -307,3 +307,18 @@
     }
   };
 
+  function setMapHeight(setHeight) {
+    if ($(window).width() > 640) {
+      if ($("ul.org-list").height() > setHeight) setHeight = $("ul.org-list").height();
```
exonian, line 312: It probably doesn't make much odds in practice in this context but re-evaluating `$("ul.org-list")` is inefficient, I think. Could be improved by setting it as a var or, in this case, going straight to setting the element height as a var.
Probably not be needed performance wise, just thought it worth mentioning as I was here, and also may make the code read easier.

cla_public/static-src/javascripts/modules/find-legal-adviser.js
```diff
@@ -307,3 +307,18 @@
...
+      $(".search-results-container").height(setHeight);
+      $(".search-results-list").height(setHeight);
```
exonian, line 314: I think these two lines could be combined, using both selectors in the lookup

cla_public/templates/checker/result/_find-legal-adviser.html
```diff
@@ -86,38 +86,38 @@ <h3 class="org-title">
...
+                      {% endfor %}
+                    </ul>
+                  {% endif %}
                 </div>
               </li>
             {% endfor %}
           </ul>
         </div>
-        <nav class="search-results-pagination">
+        <nav class="search-results-pagination" style="display:none; /*reason:this-is-dormant*/">
```
exonian, line 120: What does dormant mean here? Does it get populated and shown by ajax? Or is it currently never populated or visible in the current codebase?

cla_public/static-src/javascripts/modules/find-legal-adviser.js
```diff
@@ -307,3 +307,18 @@
     }
   };
 
+  function setMapHeight(setHeight) {
+    if ($(window).width() > 640) {
```
exonian, line 311: Would I be right in guessing that this is the breakpoint for switching from side-by-side to one-above-the-other modes? Wondering if there's a more direct way of expressing this, that's less susceptible to inadvertent changes in the future. Also worth double-checking that that point is > 640 rather than >=, just for nasty single-pixel edge effect bug avoidance.

exonian (commented):
I glanced over the css but wasn't really sure what I was looking for. Assume you're forced into all the `!important`s by overriding tightly-bound gmaps styling.

Just a few suggestions and queries on the js

cla_public/templates/checker/result/_find-legal-adviser.html
MalcolmVonMoJ: I haven't investigated thoroughly.  Pagination isn't visible, but the container is (and is always empty).  It was eating into the list space, so I added `display:none` to hide it.  I was loathed to delete it in case we need pagination added later - so I added that comment.

cla_public/static-src/javascripts/modules/find-legal-adviser.js
MalcolmVonMoJ: Done

cla_public/static-src/javascripts/modules/find-legal-adviser.js
MalcolmVonMoJ: Done

cla_public/static-src/javascripts/modules/find-legal-adviser.js
MalcolmVonMoJ: Yes and no.  The 640 value is still hard-coded into the JS file, but I have merged the two by declaring `var wideScreen = 641;` at the top and then using it in both places (changing > to >= accordingly).  

It is still susceptible to changes in the break point, but just one loose end to tie up.  

Off the back of this, I have also done a check for the 460 value, by getting the original height of the map on load, before the sizing function is run.  This will mean that a change in the SCSS variable `$map-height` will be picked up by the JS.

cla_public/templates/checker/result/_find-legal-adviser.html
exonian: If it's of no use now I'd rather delete it and trust that either it's easy enough to recreate or the devs find its removal and reinstate it from that. I don't like having non-functioning code lying around.

exonian (approved):

PR #995, opened 2020-09-02
LGA-1342 property prefixes

MalcolmVonMoJ:

Adds a property prefix so screen readers know that they are dealing with property number 1
- Sets a variable if property count > 1.
- Feeds variable into fieldset for all relevant fields.
- Radio buttons only, the currency fields pick up the fieldset legend.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/templates/macros/form.html
+++ b/cla_public/templates/macros/form.html
@@ -15,9 +15,6 @@
 #}
 {% macro fieldset(field=None, class_='govuk-fieldset ', legend=None, field_attrs={}, hidden_legend_prefix=None) %}
   {% set legend_ = field.label.text if field else legend %}
-  {% if hidden_legend_prefix %}
-    {% set legend_ = '<span class="govuk-visually-hidden">' + hidden_legend_prefix + '</span>' + legend_ | safe %}
-  {% endif %}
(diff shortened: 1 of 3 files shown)


cla_public/templates/macros/form.html
```diff
@@ -13,7 +13,7 @@
     - attrs <object> (default: None)
         Custom HTML attributes for fieldset element
 #}
-{% macro fieldset(field=None, class_='govuk-fieldset ', legend=None, field_attrs={}) %}
+{% macro fieldset(field=None, class_='govuk-fieldset ', legend=None, field_attrs={}, hidden_legend_prefix=None) %}
```
mayowaadeniyi, line 16: `hidden_legend_prefix=None`

Why is this necessary? Wouldn't the same thing happen if we left it out?

MalcolmVonMoJ: No, this is needed.  It is a new argument in the macro that we need.  We feed in a value there on the property pages telling it what the prefix needs to be.  If nothing is received, it defaults to none.

mayowaadeniyi (approved):
LGTM 👍

PR #1004, opened 2020-09-11
LGA-1339 LARP page changes

MalcolmVonMoJ:

LARP page
- If results are found, the whole page changes as opposed to just the map being shown. So the if statements now encompass most of the page rather than just the map section.  
- Most of the text from `face-to-face` has been moved into this structure in `find-legal-adviser`, after the map results bit.
- New titles, new text.
- New error messages, including tests for Crown Dependency postcodes.
- Re-wrote postcode input function to take a custom error message which forces an error style.

Intentionally left blank.

## Checklist


said-moj (approved):
👍 LGTM

PR #1006, opened 2020-09-16
LGA-1180 - Fix time selector when changing callback day

exonian:
The markup was changed when switching to the design system, breaking the
jQuery selector and therefore causing the time slots not to be updated
and a random one chosen when switching days.

At the moment, the time slots are the same on all days that have time slots (Saturdays
have no slots, rather than reduced hours), so this change won't be as apparent yet.

However, the change that broke this also caused the selector to no longer select a random
time slot when switching dates: this behaviour is also restored by this PR, and can be seen to
be working even with the current operator hours.

## Checklist


said-moj (approved):
👍 LGTM

PR #1007, opened 2020-09-17
LGA-1377 - Switch to Sentry SDK

exonian:

- Replaces Raven with the Sentry SDK, pinned to a version that's compatible with Flask 0.10
- Mostly uses the default config, but sets the `environment` tag so we can use one property for production, staging and UAT and tell them apart.
- Changes the `CLA_ENV` env var on UAT to be the name of the deployment (derived from the branch name), rather than "staging", so we can tell them apart on Sentry

Changing the `CLA_ENV` env var on UAT does change the env label that's displayed on the page, but I think this is fine.
uat

## Checklist


said-moj (approved):
👍 LGTM

PR #1009, opened 2020-09-21
LGA-1340 Pound sign prefixing for currency input fields

mayowaadeniyi:

- Let users optionally add a pound sign to any input requiring an amount to be entered 
- Rewrite validation for the front end which deals with how error classes get added. Validation on front end now matches the validation on the backend
- Write tests for the validation on frontend and backend
- Create a new test/test folder for the frontend validation called `javascript-tests/test`
- Ensure error and non-error colours are consistent for forms that require `an amount and frequency` (like pictured below) or just `an amount`

> Had to work around the fact that the £ sign is a non-ASCII character and python's default encoding is ASCII. Also on the backend, data coming in from the tests comes in byte literal while data coming from the frontend comes in unicode. I attempt to convert both to unicode format. The coveralls test shows decreased coverage because since data coming from the tests are in bytes literal, they get decoded and the `except block` never gets executed via the tests. It only gets executed when submitting input from the frontend.

**Latest fix:** Latest commits are to fix the issue pictured below making sure that if the user has entered an invalid amount and submitted, the input element will be fully red when not focusing on it and fully black when focusing on it.

image

**Worth pointing out the JS test is run using `jest`. Jest has been configured to print out coverage for the test results similar to coveralls. Possible suggestions (as they are out of scope for this ticket):**

1. Integrate the JS test into CircleCI
2. If possible integrate the JS test with coveralls or use the results produced from jest mentioned above and integrate that with CircleCI

## Checklist

### DIFF
--- a/cla_public/apps/checker/fields.py
+++ b/cla_public/apps/checker/fields.py
@@ -161,22 +161,21 @@ class MoneyField(SetZeroIntegerField):
         self.min_val = min_val
         self.max_val = max_val
 
-    def convert_amount_to_utf8(self, amount):
-        # Covert pounds side of the input into UTF-8 format
+    def extract_pounds_and_pence(self, valuelist):
+        pounds, _, pence = valuelist[0].strip().partition(".")
         try:
-            amount = amount.decode("utf-8")
+            pounds = pounds.decode("utf-8")
         except UnicodeEncodeError:
             # Input is already in UTF-8 format
             pass
 
-        return amount
+        # xa3 is the numeric character reference for the pound sign
+        pounds = re.sub(r"^\xa3|[\s,]+", "", pounds)
+        return pounds, pence
 
     def process_formdata(self, valuelist):
         if valuelist:
-            pounds, _, pence = valuelist[0].strip().partition(".")
-            pounds = self.convert_amount_to_utf8(pounds)
-            # xa3 is the numeric character reference for the pound sign
-            pounds = re.sub(r"^\xa3|[\s,]+", "", pounds)
+            pounds, pence = self.extract_pounds_and_pence(valuelist)
 
             if pence:
                 if len(pence) > 2:
--- a/cla_public/apps/checker/fields.py
+++ b/cla_public/apps/checker/fields.py
@@ -164,7 +164,7 @@ class MoneyField(SetZeroIntegerField):
     def process_formdata(self, valuelist):
         if valuelist:
             pounds, _, pence = valuelist[0].strip().partition(".")
-            pounds = re.sub(r"[£\s,]+", "", pounds)
+            pounds = re.sub(r"[\s,]+", "", pounds)
 
             if pence:
                 if len(pence) > 2:
--- a/cla_public/apps/checker/fields.py
+++ b/cla_public/apps/checker/fields.py
@@ -164,7 +164,7 @@ class MoneyField(SetZeroIntegerField):
     def process_formdata(self, valuelist):
         if valuelist:
             pounds, _, pence = valuelist[0].strip().partition(".")
-            pounds = re.sub(r"[£\s,]+", "", pounds)
+            pounds = re.sub(r"^(£|\xa3)|[\s,]+", "", pounds)
 
             if pence:
                 if len(pence) > 2:
--- a/cla_public/apps/checker/fields.py
+++ b/cla_public/apps/checker/fields.py
@@ -164,7 +164,7 @@ class MoneyField(SetZeroIntegerField):
     def process_formdata(self, valuelist):
         if valuelist:
             pounds, _, pence = valuelist[0].strip().partition(".")
-            pounds = re.sub(r"[\s,]+", "", pounds)
+            pounds = re.sub(r"[£\s,]+", "", pounds)
 
             if pence:
                 if len(pence) > 2:
--- a/cla_public/apps/checker/fields.py
+++ b/cla_public/apps/checker/fields.py
@@ -169,7 +169,7 @@ class MoneyField(SetZeroIntegerField):
             # Input is already in UTF-8 format
             pass
 
-        # xa3 is the numeric character reference for the pound sign
+        # xa3 is the ASCII character reference for the pound sign
         pounds = re.sub(r"^\xa3|[\s,]+", "", pounds)
         return pounds, pence
 
--- a/cla_public/apps/checker/fields.py
+++ b/cla_public/apps/checker/fields.py
@@ -176,7 +176,7 @@ class MoneyField(SetZeroIntegerField):
             pounds, _, pence = valuelist[0].strip().partition(".")
             pounds = self.convert_amount_to_utf8(pounds)
             # xa3 is the numeric character reference for the pound sign
-            pounds = re.sub(r"^(£|\xa3)|[\s,]+", "", pounds)
+            pounds = re.sub(r"^\xa3|[\s,]+", "", pounds)
 
             if pence:
                 if len(pence) > 2:
--- a/cla_public/apps/checker/fields.py
+++ b/cla_public/apps/checker/fields.py
@@ -161,9 +161,21 @@ class MoneyField(SetZeroIntegerField):
         self.min_val = min_val
         self.max_val = max_val
 
+    def convert_amount_to_utf8(self, amount):
+        # Covert pounds side of the input into UTF-8 format
+        try:
+            amount = amount.decode("utf-8")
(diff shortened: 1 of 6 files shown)


exonian (commented):
This is very, very good. Could we get a test of what happens when there is a pound sign after the decimal point, please? I think I know from the code and the existing tests, but it would be good to clarify it with a test case, given how many other really good test cases you have.

As for which behaviour we want, I think it would be good to get Alex's opinion on that. I think I like your very permissive approach, myself.

cla_public/apps/checker/fields.py
```diff
@@ -161,10 +161,22 @@ def __init__(self, label=None, validators=None, min_val=0, max_val=9999999999, *
         self.min_val = min_val
         self.max_val = max_val
 
+    def convert_amount_to_utf8(self, amount):
+        # Covert pounds side of the input into UTF-8 format
```
exonian, line 165: There's nothing in this method which makes it as specific as this comment suggests. It could be used for any text value, so I think the comment is somewhere between redundant and misleading. The comment would make sense (but also be redundant!) in the place this function is used.
If you wanted to document what this function does, I think it should better reflect what the function actually does, rather than its one current use case that it's not particularly specialised for.

cla_public/apps/checker/fields.py
```diff
@@ -161,10 +161,22 @@ def __init__(self, label=None, validators=None, min_val=0, max_val=9999999999, *
...
+            # Input is already in UTF-8 format
+            pass
+
+        return amount
+
     def process_formdata(self, valuelist):
         if valuelist:
             pounds, _, pence = valuelist[0].strip().partition(".")
-            pounds = re.sub(r"[\s,]+", "", pounds)
+            pounds = self.convert_amount_to_utf8(pounds)
```
exonian, line 177: From the commit message, I get the impression this was moved out to a method in order to satisfy a cyclomatic complexity check more than anything else. While I agree some of this code could be moved out, I think the simple utf-decode-unless-exception just doesn't warrant moving out, as it's plenty simple enough to encounter inline. If it truly is some reusable bit of logic, I think it'd be better suited in some "text utils" area of the codebase than still within the field class.

cla_public/apps/checker/fields.py
```diff
@@ -161,10 +161,22 @@ def __init__(self, label=None, validators=None, min_val=0, max_val=9999999999, *
         self.min_val = min_val
         self.max_val = max_val
 
+    def convert_amount_to_utf8(self, amount):
+        # Covert pounds side of the input into UTF-8 format
+        try:
+            amount = amount.decode("utf-8")
+        except UnicodeEncodeError:
+            # Input is already in UTF-8 format
```
exonian, line 169: Is this comment the right way around?

cla_public/apps/checker/fields.py
```diff
@@ -161,10 +161,22 @@ def __init__(self, label=None, validators=None, min_val=0, max_val=9999999999, *
...
+            pass
+
+        return amount
+
     def process_formdata(self, valuelist):
         if valuelist:
```
exonian, line 175: We could reduce cyclomatic complexity by switching from
```
if valuelist:
    do stuff
<implicitly returns None otherwise>
```
to
```
if not valuelist:
    return
do stuff
```

cla_public/apps/checker/fields.py
```diff
@@ -161,10 +161,22 @@ def __init__(self, label=None, validators=None, min_val=0, max_val=9999999999, *
...
             pounds, _, pence = valuelist[0].strip().partition(".")
```
exonian, line 176: I think getting pounds and pence from valuelist would work nicely as its own function. All the stripping, partitioning, decoding and regex subbing in one place, and then that function has its own meaningful purpose (rather than just a try/catch around an existing function) which could help a reader understand the code flow.

cla_public/apps/checker/fields.py
```diff
@@ -161,10 +161,22 @@ def __init__(self, label=None, validators=None, min_val=0, max_val=9999999999, *
...
-            pounds = re.sub(r"[\s,]+", "", pounds)
+            pounds = self.convert_amount_to_utf8(pounds)
+            # xa3 is the numeric character reference for the pound sign
+            pounds = re.sub(r"^(£|\xa3)|[\s,]+", "", pounds)
```
exonian, line 179: I'm not sure why we need the "£" when we do the try/catch decoding to get everything in the same encoding. When does a string reach here with a literal £ in?

exonian (changes requested):
Those tests are very, very good.
And good solve on the utf-8 stuff
Just need to hammer the method into a little more shape

cla_public/apps/checker/fields.py
mayowaadeniyi: I believe so as this is why the decoding fails

cla_public/apps/checker/fields.py
mayowaadeniyi: Since I have still kept the helper function and modified like discussed below, there's no need to do this anymore

cla_public/apps/checker/fields.py
exonian: I thought `.decode("utf-8")` meant decode *from* utf-8, not *to* utf-8, in which case I would expect it failing to mean it's no longer in utf-8

mayowaadeniyi: Ironically after doing some research `.decode("utf-8")` means to turn a string of bytes to a unicode string. It works the opposite way around to what you would except..

So when it meets a unicode string and it tries to decode it, it is already in unicode and not bytes, so it wouldn't work. Also learnt Python tries to convert (encode) the string to ASCII to help us so that we can decode it, but £ is a non-ASCII character resulting in the `UnicodeEncodeError`

cla_public/static-src/javascripts/modules/form-errors.js
```diff
@@ -143,18 +144,30 @@
...
+          inputElementWithValidEntry
+          .removeClass("govuk-input--error")
 
           $this.find(' .govuk-select').filter(function () {
             return ($(this).val())
           })
           .removeClass("govuk-select--error")
-
-          $this.find('.laa-currency .govuk-input.govuk-input--error').siblings(".laa-currency-prefix")
-            .addClass("laa-currency-prefix--error");
```
MalcolmVonMoJ, line 157: I think this still needs to be here, as the borders are black on error now.

cla_public/static-src/javascripts/modules/form-errors.js
```diff
@@ -143,18 +144,30 @@
...
+            return currencyValidation.isValidAmount($(this).val())
           })
-          .removeClass("govuk-input--error");
+
+          if(!inputElementWithValidEntry.length) {
+            $this.find(' .govuk-input')
+            .focus(function() {
+              $(".laa-currency-prefix--error")
+              .removeClass("laa-currency-prefix--error")
+            })
```
MalcolmVonMoJ, line 156: This needs to be more specific.  It is removing the red borders for other error fields on the page, as well as itself.

cla_public/static-src/javascripts/modules/currencyInputValidation.js
```diff
@@ -0,0 +1,42 @@
+// ($(this).val() && Number($(this).val()) > 0)
+
+function isValidAmount(input) {
+    var maxVal = 100000000
+    var minVal = 0
```
MalcolmVonMoJ, line 5: should these be `const`?

cla_public/static-src/javascripts/modules/currencyInputValidation.js
```diff
@@ -0,0 +1,42 @@
+// ($(this).val() && Number($(this).val()) > 0)
+
```
MalcolmVonMoJ, line 2: Unless you think this is still needed for some reason, delete it.  If you want it left, can you say why (in the comment, or another comment next to it)

MalcolmVonMoJ (changes requested):

cla_public/static-src/javascripts/modules/currencyInputValidation.js
mayowaadeniyi: Var is used because the file that requires it `form-error.js` seems to be using ES5 as when I try to use `const`, JS does not bundle correctly.

MalcolmVonMoJ:
This looks good to me now. 👍

exonian (approved):

PR #1011, opened 2020-09-22
LGA-1402 External links marked as such for screen readers only

MalcolmVonMoJ:

Added suffix to all links with `rel=external` to indicate to screen readers that it is a link to an external website.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/static-src/stylesheets/main.scss
+++ b/cla_public/static-src/stylesheets/main.scss
@@ -279,18 +279,8 @@ dl.govuk-list.govuk-list--bullet {
 }
 @media only screen {
   a[rel="external"]:after {
+    @extend .govuk-visually-hidden; 
     content:" (opens external site)";
-    position: absolute !important;
-    width: 1px !important;
-    height: 7px !important;
-    margin: 0 !important;
-    padding: 0 !important;
-    overflow: hidden !important;
-    clip: rect(0 0 0 0) !important;
-    -webkit-clip-path: inset(50%) !important;
-    clip-path: inset(50%) !important;
-    border: 0 !important;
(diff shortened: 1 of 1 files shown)


mayowaadeniyi (commented):
Could this not be done with a govuk hidden class?

MalcolmVonMoJ:
Yes, but that would involve adding it in manually for every external link.  This way it is added in by simply marking a link as external by `rel=external`, which we have already done.

The CSS comes from the `govuk-visually-hidden` class.

exonian:
As this is sass, is it possible to use some sort of `@extend` rule to get the best of both worlds? (ie we use the already-present `rel` selector but, instead of copy-pasting the style rules from the class into it, have sass extend them in)

mayowaadeniyi (commented):
Last point is 
image

Your badge name is the same name as your ticket. Was there a change that caused this because my staging badge names are just called staging? Or am I the one that's behind 😓

exonian:
@mayowaadeniyi Yep it was a change I made, when doing the Sentry work (because I wanted to use the environment value for error reporting, and wanted it to be able to show per-branch), so any branches recently started from master will have it. Said and I figured it was fine to change, and we can always make them be separate things if we really don't want the change to these badges

MalcolmVonMoJ:
Because the visually hidden CSS is pulled in from the GDS stylesheet, we cannot reference it using `@extend`.  I gave it a go anyway, just to be sure - but no... 😞

exonian:
Ah I was hoping the GDS stuff was being sass imported. Nevermind, then

mayowaadeniyi (approved):
LGTM 👍

PR #1013, opened 2020-09-23
LGA-1407 numbers

MalcolmVonMoJ:

Adds a `pattern` and `inputmode=numeric` to the integer entry fields for _**number of depentants**_ and _**number of children**_.  
Removes the same from the currency fields as per GDS guidelines.  

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/templates/checker/about.html
+++ b/cla_public/templates/checker/about.html
@@ -17,14 +17,14 @@
     {% call Form.fieldset(form.have_children) %}
       <div class="govuk-radios__conditional">
         {{ Form.group(form.num_children, '',
-            controlled_by=form.have_children, field_attrs={'class': 'govuk-input--width-4'}) }}
+            controlled_by=form.have_children, field_attrs={'class': 'govuk-input--width-4', 'inputmode':'numeric'}) }}
       </div>
     {% endcall %}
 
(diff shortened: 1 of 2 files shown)


mayowaadeniyi (commented):
Error messages won't be uniform with these changes, but I am assuming as you said in your last commit, these are the recommended changes by GDS?

image

MalcolmVonMoJ:
Yes, that's right.  The inclusion of the `pattern` element which brings up these HTML errors help get the correct keyboard for mobile devices.  I shall look into this again to be sure.

MalcolmVonMoJ:
I've added a `novalidate` at the top, which should have been there anyway. So that issue is now solved.

mayowaadeniyi (commented):
Very minor but could we change the * to a + on the controlled_by=form `pattern` attribute so it has to be a number? Or we could remove it entirely since it won't be validated anyway 😅

MalcolmVonMoJ:
I don't know - so no.  I think we should keep it as the GDS pattern.  They say:

> it's used to trigger the numeric keypad on iOS < 12.2

We could test to see whether + or something else has the same effect, but I don't see the point.

mayowaadeniyi (approved):
LGTM 👍

PR #1016, opened 2020-10-01
LGA-1446 - Ignore noisy, unimportant uwsgi errors

exonian:
Per https://github.com/getsentry/raven-python/issues/732#issuecomment-176854438
We don't want or need these to reach Sentry

Hard to demonstrate on UAT, as the incidence of this error is effectively zero, there, but as long as we can demonstrate that the site continues to work I think we can merge it and then see on production if it actually solves the problem.

## Checklist

### DIFF
--- a/docker/cla_public.ini
+++ b/docker/cla_public.ini
@@ -16,3 +16,6 @@ logformat={"process_name": "uwsgi", "timestamp_msec": %(tmsecs), "method": "%(me
 post-buffering = 1
 buffer-size=32768
 post-buffering-bufsize=32768
+ignore-sigpipe = true
+ignore-write-errors = true
+disable-write-exception = true


heidar (approved):
Bit of a relief knowing these aren't serious, LGTM.

exonian:
@heidar It always makes me incredibly nervous ignoring anything on the basis of it "not being serious" but I think based on other people's experience, and the absence of anyone objecting on those threads, we're ok here  😰

heidar:
> @heidar It always makes me incredibly nervous ignoring anything on the basis of it "not being serious" but I think based on other people's experience, and the absence of anyone objecting on those threads, we're ok here 😰

Yeah I had the same reaction! People in that issue you linked seemed pretty confident.

PR #1024, opened 2020-10-14
LGA-1475 Two H1s on one page

MalcolmVonMoJ:

Checked for title attribute - if set, that means a title is already used (in laalaa.html).  Otherwise use title as normal.

Intentionally left blank.

## Checklist


said-moj (approved):
👍  LGTM

PR #1027, opened 2020-10-26
LGA-1479 Erratic Other Language Select Class

MalcolmVonMoJ:

- Adds in a width class to the other language select field:
  - This fixes the still-unexplained bug.
  - This also maintains the wide breadth of the select field, increasing click area.
- Added in note for any future development on this, so the bug isn't accidentally re-introduced.

Intentionally left blank.

## Checklist


exonian (approved):
This seems sensible to me. I like the warning comment and the reference to the ticket.
Obviously never thrilled to not fix something at the source, but this looks like a good way of dealing with not being able to do that.

PR #1028, opened 2020-10-26
SPIKE LGA-1354 Install and add webhint to cirleci pipeline

mayowaadeniyi:

- Configure CircleCI config.yaml file and create a new webhint job, so webhint can be added to CircleCI pipeline
- Integrate webhint and see if it works/catches anything

> Pull request closed because this was a spike. Webhint will be fully implemented in the following ticket: **LGA-1509**

Some of the things I tested webhint for:

- Contrast
- Fieldset element without a legend (useful for screen readers)
- Tab index (just checks if a element has a tab index > 0 which means it is out of flow of the normal document layout)
- List `<li>` elements not having a `<ul>` or `<ol>` tag
- Form input elements not having a label linked to them
- Image  elements not have any alt text
- Buttons and links not having any text inside but other content e.g. an image or + symbol 

**Webhint caught 6/7 of these checks**. The only thing it didn’t catch was the fieldset without a legend check.

**Overall webhint is good for catching basic accessibility issues**. After researching the most common accessibility issues, I learnt that contrast, missing alt text on images and missing link text are among the most common. All of which webhint picks up.

Link to webhint results with all the issues above mentioned: https://13396-17939582-gh.circle-artifacts.com/0/~/project/hint-report/https-lga-1354-install-webhint-staging-checklegalaid-service-gov-uk-accessibility-statement.html

(_All issues found in the accessibility column in the link above_)

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,7 +119,7 @@ jobs:
           name: Install and run webhint
           command: | 
             npm install --no-save hint
-            node node_modules/hint/dist/src/bin/hint.js https://$RELEASE_HOST/start
+            node node_modules/hint/dist/src/bin/hint.js https://$RELEASE_HOST/
 
       - store_artifacts:
           path: ~/project/hint-report/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,7 +107,7 @@ jobs:
           name: Install and run webhint
           command: | 
             npm install --no-save hint puppeteer
-            
+            node node_modules/hint/dist/src/bin/hint.js https://staging.checklegalaid.service.gov.uk/scope/diagnosis/
   test:
     docker:
       - image: circleci/python:2.7
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,6 +107,10 @@ jobs:
       - checkout
       - puppeteer/install
       - run: 
+          name: Delete existing javascript packages
+          command: | 
+            rm -rf node_modules package.json package-lock.json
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -98,6 +98,22 @@ jobs:
           command: |
             source py3-lint-env/bin/activate
             black --check --diff cla_public
+  webhint:
+    docker:
+      - image: circleci/python:latest-node-browsers
```
exonian, line 103: I would start with the node image. We don't need to actually run our Python code so I don't think we need the Python image, and using the node one gets us ready to install hint much more easily. Unless we can't stop it installing from our project's package.json, but I think we should be able

.circleci/config.yml
```diff
@@ -98,6 +98,16 @@ jobs:
           command: |
             source py3-lint-env/bin/activate
             black --check --diff cla_public
+  webhint:
+    docker:
+      - image: node:15.0.0-stretch
```
exonian, line 103: Sorry. Me again! We were going to try one of the official `circle/` images, weren't we? Cos we figured that would be more ready to get working here (it shouldn't have any of the perms issues we had fun solving last week, for example)

exonian:
Maybe I was on the wrong track with switching from the circle python node image to the circle node one, after all. I thought it was that because I could see it running our app's `package.json` stuff, which isn't going to work on the same version of node as hint will work, but I think it wasn't necessarily having the wrong image that was the cause

PR #1029, opened 2020-11-09
LGA-1508 Margin for postcode input

MalcolmVonMoJ:

Added a class to the postcode input that puts a margin beneath it, so the search button doesn't butt up against it in mobile view.  

Added in a variable for the CSS file - to jolt any return users to refresh their CSS.

## Checklist


mayowaadeniyi (approved):
LGTM 👍

PR #1034, opened 2020-11-26
LGA-1538 erratic drop-down fields on contact page

MalcolmVonMoJ:

Defines a class attribute for all drop-downs.  

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/templates/macros/subform.html
+++ b/cla_public/templates/macros/subform.html
@@ -43,7 +43,6 @@
     </ul>
     {{ Form.group(form.thirdparty.full_name, field_attrs={'class': 'm-large'}) }}
     {{ Form.group(form.thirdparty.relationship,
-      '',
       field_attrs={'class': ''}) }}
     {{ Form.group(form.thirdparty.contact_number, field_attrs={'class': 'm-large', 'autocomplete': 'tel', 'type': 'tel'}) }}
     {{ time_picker(form.thirdparty.time) }}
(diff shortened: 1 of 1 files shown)


cla_public/templates/macros/subform.html
```diff
@@ -42,7 +42,9 @@
       <li>{{ _('If your nominated person misses the call you’ll need to complete this form again.') }}</li>
     </ul>
     {{ Form.group(form.thirdparty.full_name, field_attrs={'class': 'm-large'}) }}
-    {{ Form.group(form.thirdparty.relationship) }}
+    {{ Form.group(form.thirdparty.relationship,
+      '',
```
exonian, line 46: What is this empty argument? I can see it in many but not all (eg `form.thirdparty.contact_number` directly below) instances

MalcolmVonMoJ: Not 100% sure.  I just copied the pattern, including the empty argument.  I could see element classes were passed into it in some instances.  

Looking into if further, this passes extra classes into the group element, such as the conditional element for show-hide things.  

I am surprised to see it declared null, like you obviously are, but I just accepted it and carried on.

MalcolmVonMoJ: I've just tried out removing it and it seems to work fine.  I can't see any difference in the rendered HTML.  I shall remove the remaining ones.

cla_public/templates/macros/subform.html
```diff
@@ -42,7 +42,8 @@
       <li>{{ _('If your nominated person misses the call you’ll need to complete this form again.') }}</li>
     </ul>
     {{ Form.group(form.thirdparty.full_name, field_attrs={'class': 'm-large'}) }}
-    {{ Form.group(form.thirdparty.relationship) }}
+    {{ Form.group(form.thirdparty.relationship,
+      field_attrs={'class': ''}) }}
```
mayowaadeniyi, line 46: Can we remove the linespace so it's on one line? Just so it's a bit more readable 😅

mayowaadeniyi (approved):
LGTM 👍

PR #1047, opened 2021-01-06
LGA-1574 - Reinstate noisy, but potentially important uwsgi errors

exonian:

Stops filtering out these types of i/o errors from reaching Sentry. These errors may actually have (at least partly) indicated a problem after all. We've added config to cla_backend to try to reduce them, so I'd like to see if that's actually had any effect on them.

Intentionally left blank.

## Checklist


said-moj (approved):
👍 LGTM

PR #1048, opened 2021-01-11
LGA-1570 LARP Modern Slavery category

MalcolmVonMoJ:

- New category for Modern Slavery and Human Trafficking
- New function for analysing the codes of previous pages in new `utils.py` file
  - Adapted from the existing QE check

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/apps/checker/constants.py
+++ b/cla_public/apps/checker/constants.py
@@ -175,6 +175,7 @@ LAALAA_PROVIDER_CATEGORIES_MAP = {
     "other": ["other"],
 }
 
+# adds another to the above - see get_category_for_larp
 LAALAA_PROVIDER_CATEGORIES_MAP.update({"traffickingslavery": ["mosl"]})
 
 END_SERVICE_FLASH_MESSAGE = _(
--- a/cla_public/apps/checker/constants.py
+++ b/cla_public/apps/checker/constants.py
@@ -24,12 +24,12 @@ CATEGORIES = [
     (
         "discrimination",
         _(u"Discrimination"),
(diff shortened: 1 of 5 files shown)


cla_public/apps/scope/views.py
```diff
@@ -71,6 +72,13 @@ def add_link(choice):
 
         return render_template("scope/diagnosis.html", choices=display_choices, nodes=nodes)
 
+    def get_category_for_larp(self, session):
+        category_for_LARP = session.checker.category
```
exonian, line 76: I would move this into an `else` statement on the `if`, so that we're only assigning this value if there's no category match, rather than setting it and then maybe overloading it

exonian (changes requested):

cla_public/apps/checker/constants.py
```diff
@@ -132,6 +132,7 @@
     "publiclaw",
     "aap",
     "employment",
+    "traffickingslavery",
```
said-moj, line 135: Where does `traffickingslavery` come from?
does it come from the backend?

cla_public/apps/scope/views.py
MalcolmVonMoJ: 👍

cla_public/apps/checker/constants.py
MalcolmVonMoJ: Yeah, this isn't needed. Removed. Thanks.

exonian (approved):
Obviously just make sure to UAT it for both the new functionality and the quick exit stuff, as we don't have tests I'd 100% rely on :)

PR #1051, opened 2021-01-14
LGA-1585 Modern Slavery LARP to search all categories

MalcolmVonMoJ:

Comments out old code and makes `traffickingslavery` search for a blank, which returns all providers.  
Old code to be reinstated when new `mosl` category is set up on Production.

Intentionally left blank.

## Checklist


exonian (changes requested):
Nice fix but just make the change, don't leave anything commented out.
We'll just revert this change when we're ready, so we don't need the code sat there commented out really.

I'd also name the PR (which affects the merge commit) a bit more descriptively of what it does, maybe.

exonian (approved):
Thanks!

PR #1052, opened 2021-01-15
LGA-1586 alt help opening hours in welsh

MalcolmVonMoJ:

- Adds translation tag around opening times field allowing translation.
- Adds new CSS class for styling multi-line opening times (as the Welsh translation preserves line-returns).
- Adds Welsh translations for all normal opening times
  - Not including a few language-heavy opening times that need translations

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/static-src/stylesheets/_help-organisations.scss
+++ b/cla_public/static-src/stylesheets/_help-organisations.scss
@@ -42,6 +42,7 @@
       font-weight: bold;
       display: inline-block;
       margin-right: 5px;
+      vertical-align: top;
     }
 
     dd {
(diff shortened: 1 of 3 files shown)


exonian:
I don't think this is a particularly appropriate solution. Maybe there's no better alternatives and it needs to be done right now so it needs to be done like this, but I can't even see where this design was decided upon or any considerations weighed up.

MalcolmVonMoJ:
I agree this method isn't ideal.  But with no consistency in the English I didn't see how else to approach this.  This is very much a "for now" solution.  When your ticket is done it will enable us to maintain the English, which should allow us to have a better approach for the Welsh.  

The Welsh now has more consistency than the English does, as all the variations of 9-5 (hyphens, to's, en dashes etc) are translated in the same way.  

Perhaps, after we rationalise the English, this could be revisited to be done in a more modular way.

exonian (approved):
As discussed elsewhere, I think there were better approaches open to us but given that this work's been done like this, it's been done well and the only risk of its brittleness is that we end up back where we are now for a particular provider, so I think we're best off merging this even if we end up replacing it before too long

PR #1054, opened 2021-01-25
LGA-1509 Test happy path with webhint

mayowaadeniyi:

- Runs webhint against every page in happy path journey
- Use recommended webhint config for scanning local files
- Upload webhint results to artifacts on circle
- Configure webhint so it uses _your_ UAT URL and **fix current failing of webhint on circle due to making webhint use my old UAT URL**

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,7 +126,8 @@ jobs:
           name: Install and run webhint
           command: | 
             npm install --no-save hint
-            node node_modules/hint/dist/src/bin/hint.js cypress/pages/
+            cd ./cypress/pages/
+            node ../../node_modules/hint/dist/src/bin/hint.js ./
       - store_artifacts:
           path: ~/project/hint-report/
   test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,13 +124,9 @@ jobs:
             node node_modules/.bin/cypress run --env UAT_URL=https://$RELEASE_HOST/start --spec "cypress/integration/happy-path.spec.js"
       - run: 
           name: Install and run webhint
-          # Move .hintrc into cypress/pages folder so webhint uses our custom webhint config
           command: | 
             npm install --no-save hint
-            cp .hintrc cypress/pages/
-            cd ./cypress/pages/
-            node ../../node_modules/hint/dist/src/bin/hint.js ./
-            mv hint-report ../..
+            node ~/project/node_modules/hint/dist/src/bin/hint.js cypress/pages/
       - store_artifacts:
           path: ~/project/hint-report/
   test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,13 +122,13 @@ jobs:
             npm i cypress-downloadfile
             mkdir ~/project/cypress/pages
             node node_modules/.bin/cypress run --spec "cypress/integration/happy-path.spec.js"
-      # - run: 
-      #     name: Install and run webhint
-      #     command: | 
-      #       npm install --no-save hint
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -122,13 +122,15 @@ jobs:
...
-      #     path: ~/project/hint-report/
+      - run: 
+          name: Install and run webhint
+          command: | 
+            npm install --no-save hint
+            cd ./cypress/pages/
+            node ../../node_modules/hint/dist/src/bin/hint.js ./
+            mv hint-report ../..
+      - store_artifacts:
+          path: ~/project/hint-report/
```
exonian, line 133: Is there a reason that leaving the report where it is and specifying its path in `store_artifacts` doesn't work? That way strikes me as neater

cypress/integration/happy-path.spec.js
```diff
@@ -1,7 +1,7 @@
 describe('Happy path testing', function() {
   it('Should complete happy path successfully from scope diagnosis to confirmation page', function () {
 
-    cy.visit("https://lga-1543-install-and-integra.staging.checklegalaid.service.gov.uk/start")
+    cy.visit("https://lga-1509-test-happy-path-wit.staging.checklegalaid.service.gov.uk/start")
```
exonian, line 4: Why is this hard-coded? Is this going to only run on master branch, and therefore staging url?
I had assumed it would be used on feature branches, so would need to use the UAT url dynamically

exonian (changes requested):
I think it being all one report is absolutely fine for now. Out of interest, do you happen to know if it would be simple to change that if we were to choose to do so?

Couple of potential changes to make, though

.circleci/config.yml
mayowaadeniyi: Ah ok no there are no issues with that. I did it this way simply because I thought this was neater 😂 as well as the fact that the hint-report is not associated with cypress and wanting to leave the `store_artifacts` the same

cypress/integration/happy-path.spec.js
mayowaadeniyi: Yeah this is one of the bugs I need to fix. As part of the ticket, I will make this dynamic by using environmental variables and running the command in circleci with the correct cli arguments

.circleci/config.yml
exonian: OK. I don't see the rationale for not wanting to change the `store_artifacts` path but having it end up not in the cypress files makes some sense to me. I had expected there to be config to specify the destination for the report but I can't find any so moving it afterwards is fine.
Do we have to run webhint from where we do? Could we instead cd to where we want the report to be (`~/project/`) and run it there, with a path to the pages instead of `.`?

cypress/integration/happy-path.spec.js
exonian: OK

.circleci/config.yml
mayowaadeniyi: I added the reason why in the description of one of my commits. I'll see if there is anyway way around it. Here's what I wrote:

_Webhint has a special command (./) used at the end for scanning an entire folder. When I tried added further folders to the command e.g. ./cypress/pages, it would error out.

The command seems be specific and only work when your in the folder to be scanned._

exonian: Sorry, didn't see this until just now. I would wager you just might be in the wrong place for the path you used, to be honest

mayowaadeniyi: Turns out you was very right! (as per usual)

exonian (approved):

PR #1059, opened 2021-02-18
LGA-1570 Adding route for larp modern slavery

said-moj:

Adding route for larp modern slavery

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/apps/checker/constants.py
+++ b/cla_public/apps/checker/constants.py
@@ -176,7 +176,7 @@ LAALAA_PROVIDER_CATEGORIES_MAP = {
 }
 
 # adds another to the above - see get_category_for_larp
-LAALAA_PROVIDER_CATEGORIES_MAP.update({"traffickingslavery": [""]})
+LAALAA_PROVIDER_CATEGORIES_MAP.update({"traffickingslavery": ["mosl"]})
 
 END_SERVICE_FLASH_MESSAGE = _(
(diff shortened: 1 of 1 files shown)


cla_public/apps/checker/constants.py
```diff
@@ -176,7 +176,7 @@
 }
 
 # adds another to the above - see get_category_for_larp
-LAALAA_PROVIDER_CATEGORIES_MAP.update({"traffickingslavery": [""]})
+LAALAA_PROVIDER_CATEGORIES_MAP.update({"traffickingslavery": ["mosl"]})
```
exonian, line 179: Is there a reason this isn't just included in the above dict literal?
I assume @MalcolmVonMoJ had his reasons, but I can't figure it out

said-moj: I think I asked malcom to add it on a separate line on one of his PR. My reason was that all of the above ones were direct mappings between LAALAA categories and CLA categories.

However traffickingslavery is like a psuedo category being generated by `cla_public.apps.scope.views.ScopeDiagnosis.get_category_for_larp`

exonian: Ah yes that makes sense. Let's leave it as is and maybe if we add more in future we can move to defining two dicts and then combining them

exonian (approved):

PR #1064, opened 2021-03-11
LGA 1636 - Display full summary

Pheonnexx:

This fixes the defect that stop someone changing data that would have made them ineligible.  

> For example a person mistakenly enters 100,000 for their wage which then skips to the review page.  However when on the review page it didn't display the data that was entered, and then they wouldn't know to go back and correct it.

I reintroduced the use review_steps method which had never been removed from the code.

Intentionally left blank.

## Checklist


exonian (approved):
Code looks good. Very convenient it had been left there unused 😅

PR #1068, opened 2021-03-22
LGA-1631 Message Switch

Pheonnexx:

This pull request is to:

- prepare for work in LGA 1631.  Using config mapping to have a variable that is accessible to use to amend quickly. 

 - amend the way that environment variables are brought in by Kubernetes in cla_public.  This allows us to align with the other repos such as cla_frontend.

Amends the helpers, values for each environment and deployment.yaml
Changes the way enVars are set.
Removes Raven enVars no longer used.

There is documentation which tells you exactly how to update the env var in the config map however for reference it revolves around the command `kubectl edit configmap emergency-message`

## Checklist


helm_deploy/cla-public/values-production.yaml
```diff
@@ -5,11 +5,6 @@
 image:
   pullPolicy: IfNotPresent
 
-environment: production
```
mayowaadeniyi, line 8: Do we want to remove this?

helm_deploy/cla-public/templates/_helpers.tpl
```diff
@@ -61,3 +61,37 @@ Create the name of the service account to use
...
+  value: "{{ $data.value }}"
+{{- else if $data.secret }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ $data.secret.name }}
+      key: {{ $data.secret.key }}
+      {{- if eq $environment "development" }}
+      optional: true
+      {{- else }}
+      optional: {{ $data.secret.optional | default false }}
```
mayowaadeniyi, line 83: Since we have removed `useDevValues` from each yaml file, can we make remove `$data.secret.optional` since it is no longer being used and make it just be false ?

Same with line 93

exonian: `useDevValues` was a way of saying "true if on dev, otherwise false"
But if we made your proposed change, we'd be locked in to having every env var set like that, and not be able to make any optional in non-dev environments

mayowaadeniyi: I thought with that logic in mind, we would want it to always be false if not on dev? Plus we aren't setting an optional key in any of our secrets which is why I thought to remove it.

helm_deploy/cla-public/templates/_helpers.tpl
```diff
@@ -61,3 +61,37 @@ Create the name of the service account to use
     {{ default "default" .Values.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
+
```
mayowaadeniyi, line 64: Can we add a comment here on what this does like on line 54 - 56 e.g.

`{{/*
Create the name of the service account to use
*/}}`

helm_deploy/cla-public/templates/_helpers.tpl
exonian: Lines 80-81 mean that on dev it's always optional
Lines 82-83 are for non-dev, so what we do in lines 80-81 isn't relevant to those

You're right we don't currently have any optional secrets on non-dev. But we _could_ set the optional key on our secrets, and I'd prefer a helper that's ready for that as we have it. Making the change you suggest means we'd have to update the helper if we ever wanted that, which personally I don't think is worth it

helm_deploy/cla-public/templates/_helpers.tpl
exonian: Good catch. Do you want to go see if cla_backend (where we shamelessly stole this from) has such a thing?
And then either suggest that here, or suggest some wording of your own

Pheonnexx: I put a description up, can you check if its suitable @mayowaadeniyi

helm_deploy/cla-public/templates/_helpers.tpl
Pheonnexx: Going to keep as is!

helm_deploy/cla-public/values-production.yaml
Pheonnexx: Nice catch and re added!

helm_deploy/cla-public/templates/_helpers.tpl
mayowaadeniyi: Could we say `Takes environment variables set in the values and values-environment file, and arranges them in an appropriate format`?

said-moj: Did you leave `What does this pull request do` and `Any other changes that would benefit highlighting?` blank and put your message outside of those headers intentionally?

Also can you mention how to turn on emergency message? (is there a kubectl command?)

Pheonnexx: I see what I did there @said-moj  and have corrected that, I have instructions ready to go into confluence for this, but also put the command to edit the config map into the description.

helm_deploy/cla-public/values-staging.yaml
```diff
@@ -18,13 +15,23 @@ ingress:
 
 # Lists don't deep merge, so this list of envVars overrides anything defined in an earlier values file
```
exonian, line 16: This is no longer true, right?

Pheonnexx: unneeded comments have now been removed.

cla_public/templates/privacy.html
```diff
@@ -152,7 +150,7 @@ <h2 class="govuk-heading-m">{{ _('Retention period for information collected') }
     {% trans %}Your personal information will not be retained for any longer than is necessary for the lawful purposes for which it has been collected and processed. This is to ensure that your personal information does not become inaccurate, out of date or irrelevant. The Legal Aid Agency have set retention periods for the personal information that we collect, this can be accessed via our website:{% endtrans %}
   </p> 
   <p class="govuk-body">
-    {% trans %} <a class="govuk-link" href="https://www.gov.uk/government/publications/record-retention-and-disposition-schedules" rel="external">See our record retention and disposition schedules</a>{% endtrans %}
+   <a class="govuk-link" href="https://www.gov.uk/government/publications/record-retention-and-disposition-schedules" rel="external">{% trans %}See our record retention and disposition schedules{% endtrans %}</a>
```
exonian, line 153: I'm a bit suspicious of this change. Is it intended?

exonian: This was spurious diffing

exonian (approved):

PR #1081, opened 2021-05-05
LGA-989 - Restart session if its not already available when giving feedback

said-moj:

Restart session if its not already available when giving feedback

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_public/apps/base/views.py
+++ b/cla_public/apps/base/views.py
@@ -133,6 +133,12 @@ class ReasonsForContacting(AbstractFeedbackView):
         referrer = re.sub(r"^(.*:)//([A-Za-z0-9-.]+)(:[0-9]+)?/", "/", request.referrer or "")
         return render_template(self.template, form=self.form, referrer=referrer)
 
+    def get(self, *args, **kwargs):
+        if not session or not session.is_current:
+            session.checker["started"] = datetime.datetime.now()
+
+        return super(ReasonsForContacting, self).get(*args, **kwargs)
+
     def post(self):
         if self.form.validate_on_submit():
             if len(self.form.reasons.data) == 0:


cla_public/apps/base/views.py
```diff
@@ -133,6 +133,12 @@ def render_form(self):
         referrer = re.sub(r"^(.*:)//([A-Za-z0-9-.]+)(:[0-9]+)?/", "/", request.referrer or "")
         return render_template(self.template, form=self.form, referrer=referrer)
 
+    def get(self, *args, **kwargs):
```
exonian, line 136: This looks as though it would work perfectly well.

However, we do have [`EnsureSessionExists`](https://github.com/ministryofjustice/cla_public/blob/432bdfb7244ccc22035846a5350890ac068c21a1/cla_public/libs/views.py#L14) (that I found when searching for `session.is_current`), that we can subclass (eg `class ReasonsForContacting(EnsureSessionExists, AbstractFeedbackView)`)

The `get` and `post` methods here get called by the `dispatch_request` of `MethodView` (that we inherit from via `AbstractFeedbackView`) so as long as we put `EnsureSessionExists` first, it will be ahead of it in the MRO (method resolution order) and the execution will go `dispatch_request` from `EnsureSessionExists`, which will call `super()`, which will then reach `dispatch_request` from `MethodView`, which will then call `get()` or `post()` appropriately.

It's honestly not a big difference, and if we didn't already have that mixin class I wouldn't suggest writing it for this, but as we have it I think I'd like to use it, assuming it all works in the way I think it does.

exonian (approved):
I would rather have this released than not. I would like us to consider the code refactor I suggested, but the functionality has been tested and as this directly improves the experience for the public I can't justify not merging and releasing it, and we can revisit it later.

PR #1114, opened 2021-12-01
LGA-1914 - Remove setup kube auth

exonian:

Removes the use of the `setup-kube-auth` script
Specifies additional contexts so that each step can have only the kubernetes env vars it needs
Creates and uses a script to authenticate with the kubernetes cluster given the presence of these env vars

The staging_deploy and production_deploy steps could be consolidated but it's not strictly in the scope of this and I'm pressed for time so I left it for now

## Checklist


said-moj (approved):
👍 LGTM

PR #1185, opened 2023-05-10
LGA-2423 ITHC update images ADDITION

jogilder:
Now both match more up to date image used for staging - this means that the slack notification works because the certificate is up to date.

Required.

Intentionally left blank.

## Checklist


TawneeOwl (approved):
Looks good without the package-lock file

PR #1191, opened 2023-06-15
LGA-2454 Add Welsh Translations for Notify Emails

jogilder:

We previously used Sendgrid to send emails and have moved to using Notify. These emails are only sent in English.

Using Sendgrid, the email text were translated and then sent over. This is not possible in Notify, the template needs to be in Welsh.

Mirrors the original code as it does not send welsh email if you have ticked welsh on the form but haven’t chosen to view pages in welsh.

I haven't added any tests as I figured there are already tests for checking the locale. Can add if you think it is worth it.

I believe I have added in the Welsh translations. Someone will need to check those though...

Mirrors the original code as it does not send welsh email if you have ticked welsh on the form but haven’t chosen to view pages in welsh.

Need to make a decision on whether to keep the old templates and references to SMTP server in.

## Checklist


Pheonnexx (approved):
LGTM.

PR #1219, opened 2023-11-28
Amend jquery version

TawneeOwl:

Updates Jquery to version 3.7.1

Intentionally left blank.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "gulp-rename": "~1.2.0",
     "gulp-sass": "^4.0.1",
     "gulp-uglify": "^3.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "gulp-rename": "~1.2.0",
     "gulp-sass": "^4.0.1",
     "gulp-uglify": "^3.0.0",
+    "jquery": "3.7.1",
```
psweeting1, line 29: This will not make the project use that type of jQuery. Also it appears that this was not in the package file originally, which tells me it's working with Jquery in a different way.

PR #1247, opened 2024-03-21
LGA-3022 Track custom errors on GA

digitalali-moj:
Ticket: https://dsdmoj.atlassian.net/browse/LGA-3022

Enables GA on error pages so we can see what happened within a couple of days and count how many people were affected. 

1. Adding some error codes that were missing before
2. Giving original 4xx template a bit more generic title
4. Adding Google Analytics tracking to both 4xx and 5xx error templates. Includes PageView and event tracking. Benefit of the event is to easily see stats and categorise errors on GA by code. Technically not needed but convenience.
5. Inserting error code for maintenance page into template otherwise it's blank

Maintenance page (/maintenance) - this uses 5xx template. 

The code used is universal GA code in line with the rest of the app. The new app should be using Tag Man and GA4 for tracking events. I don't personally see the need to factorise this code as I think it would increase complexity, but happy to take on suggestions.

Note: I added a couple of 4xx codes to help with security mainly. I also added 400. The original template was a bad request template but didn't seem to be used for that. I changed the template to make it a bit more generic for the user.

QUESTION: Why was 400 and 403 code not in the list of error codes before? Could there some AJaXy reason? I couldn't see any but thought to double check.

## Checklist


BenMillar-CGI (approved):
LGTM, great stuff! 👍

Pheonnexx (approved):
LGTM

PR #1262, opened 2024-05-09
LGA-2991 - Disable indexing and whitelisting of Non-Production environments

said-moj:

Disable indexing and whitelisting of Non-Production environments

Pingdom IP addresses also needed to be added to the whitelist. However these frequently change, so an additional script is added to discover these IP addresses.

## Checklist

### DIFF
--- a/bin/staging_multideploy.sh
+++ b/bin/staging_multideploy.sh
@@ -17,5 +17,6 @@ helm upgrade $CLEANED_BRANCH_NAME \
   --set image.repository=$DOCKER_REPOSITORY \
   --set image.tag=$IMAGE_TAG \
   --set dashboard.enabled=false \
+  --set-string pingdomIPs=$PINGDOM_IPS \
   --force \
   --install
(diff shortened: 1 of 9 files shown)


Pheonnexx (dismissed):

bin/staging_multideploy.sh
```diff
@@ -17,5 +17,6 @@ helm upgrade $CLEANED_BRANCH_NAME \
   --set image.repository=$DOCKER_REPOSITORY \
   --set image.tag=$IMAGE_TAG \
   --set dashboard.enabled=false \
+  --set-string pingdomIPs=$PINGDOM_IPS \
```
digitalali-moj, line 20: Is this dynamic staging. If so do we need pingdom access? Is it simpler to keep?

digitalali-moj (approved):
Excellent work! One little question

TawneeOwl (approved):
LGTM

bin/staging_multideploy.sh
said-moj: yes this dynamic staging and no we don't need pingdom access on this environment. However it makes writing the config writing simpler

Pheonnexx (approved):

PR #1275, opened 2024-08-12
LGA-3200 - Reset the contact phone number when another contact method is selected

said-moj:

Reset the contact phone number when another contact method is selected.

Intentionally left blank.

## Checklist


digitalali-moj (approved):
Thanks looks good. 

Side questions - do we need to clear any other fields other than all phone numbers when you toggle the radios?

BenMillar-MOJ (approved):
LGTM, thanks for this!

Works as expected on staging.

said-moj:
> Thanks looks good.
> 
> Side questions - do we need to clear any other fields other than all phone numbers when you toggle the radios?

Yes I think it would be good to clear all other fields but I just limited my change to the scope of the ticket as I am a bit worried about breaking anything.

PR #1284, opened 2024-11-13
LGA-3339 - Fix broken add property button

said-moj:

Fix broken add property button

The promise returned from $.post no longer has a `success `method. Instead the `done` method should be used

From the jquery docs:
> Deprecation Notice
The jqXHR.success(), jqXHR.error(), and jqXHR.complete() callback methods are removed as of jQuery 3.0. You can use jqXHR.done(), jqXHR.fail(), and jqXHR.always() instead.

The bug was probably introduced around the same time that the jquery version was upgraded in #1220 

## Checklist

### DIFF
--- a/cla_public/static-src/javascripts/modules/properties.js
+++ b/cla_public/static-src/javascripts/modules/properties.js
@@ -37,7 +37,7 @@
       $.post('', this._injectFormData({
         name: button.name,
         value: button.value
-      })).success(function(res) {
+      })).done(function(res) {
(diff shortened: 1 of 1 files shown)


digitalali-moj (approved):
LGTM, cheers!

BenMillar-MOJ (approved):
👍 That was a lot easier than expected!

TawneeOwl (approved):
LGTM

